PR #10, opened 2021-06-23
Add Initial elasticsearch model

jolynch:
Primary: @abersnaze 

Feedback welcome from anybody on the ES team, especially in the FIXMEs.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/elasticsearch.py
+++ b/service_capacity_modeling/models/org/netflix/elasticsearch.py
@@ -15,11 +15,9 @@ from service_capacity_modeling.interface import CapacityRequirement
 from service_capacity_modeling.interface import certain_float
 from service_capacity_modeling.interface import certain_int
 from service_capacity_modeling.interface import Clusters
-from service_capacity_modeling.interface import Consistency
 from service_capacity_modeling.interface import DataShape
 from service_capacity_modeling.interface import Drive
 from service_capacity_modeling.interface import FixedInterval
-from service_capacity_modeling.interface import GlobalConsistency
 from service_capacity_modeling.interface import Instance
 from service_capacity_modeling.interface import Interval
 from service_capacity_modeling.interface import QueryPattern
@@ -36,17 +34,31 @@ from service_capacity_modeling.stats import dist_for_interval
 logger = logging.getLogger(__name__)
 
 
-# Pebble does Leveled compaction with tieres of size??
-# (FIXME) What does pebble actually do
-def _crdb_io_per_read(node_size_gib, sstable_size_mb=1000):
-    gb = node_size_gib * 1024
-    sstables = max(1, gb // sstable_size_mb)
-    # 10 sstables per level, plus 1 for L0 (avg)
-    levels = 1 + int(math.ceil(math.log(sstables, 10)))
+def _target_rf(desires: CapacityDesires, user_copies: Optional[int]) -> int:
+    if user_copies is not None:
+        assert user_copies > 1
+        return user_copies
+
+    # Due to the relaxed durability and consistency requirements we can
+    # run with RF=2
+    if desires.data_shape.durability_slo_order.mid < 1000:
+        return 2
+    return 3
+
+
+# Looks like Elasticsearch (Lucene) uses a tiered merge strategy of 10
+# segments of 512 megs per
+# https://lucene.apache.org/core/8_1_0/core/org/apache/lucene/index/TieredMergePolicy.html#setSegmentsPerTier(double)
+# (FIXME) Verify what elastic merge actually does
+def _es_io_per_read(node_size_gib, segment_size_mb=512):
+    size_mib = node_size_gib * 1024
+    segments = max(1, size_mib // segment_size_mb)
+    # 10 segments per tier, plus 1 for L0 (avg)
+    levels = 1 + int(math.ceil(math.log(segments, 10)))
     return levels
 
 
-def _estimate_cockroachdb_requirement(
+def _estimate_elasticsearch_requirement(
     instance: Instance,
     desires: CapacityDesires,
     working_set: float,
@@ -60,15 +72,15 @@ def _estimate_cockroachdb_requirement(
     The input desires should be the **regional** desire, and this function will
     return the zonal capacity requirement
     """
-    # Keep half of the cores free for background work (compaction, backup, repair)
-    needed_cores = sqrt_staffed_cores(desires) * 2
+    # Keep half of the cores free for background work (merging mostly)
+    needed_cores = math.ceil(sqrt_staffed_cores(desires) * 1.5)
     # Keep half of the bandwidth available for backup
     needed_network_mbps = simple_network_mbps(desires) * 2
 
     needed_disk = math.ceil(
         (1.0 / desires.data_shape.estimated_compression_ratio.mid)
         * desires.data_shape.estimated_state_size_gib.mid
-        * copies_per_region
+        * copies_per_region,
     )
 
     # Rough estimate of how many instances we would need just for the the CPU
@@ -82,7 +94,7 @@ def _estimate_cockroachdb_requirement(
     # hitting disk per instance. If we don't have many reads we don't need to
     # hold much data in memory.
     instance_rps = max(1, reads_per_second // rough_count)
-    disk_rps = instance_rps * _crdb_io_per_read(max(1, needed_disk // rough_count))
+    disk_rps = instance_rps * _es_io_per_read(max(1, needed_disk // rough_count))
     rps_working_set = min(1.0, disk_rps / max_rps_to_disk)
 
     # If disk RPS will be smaller than our target because there are no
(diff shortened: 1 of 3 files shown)


service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+        and consistency is not None
+        and consistency.target_consistency != AccessConsistency.read_your_writes
+    ):
+        return 2
+    return 3
+
+
+# Looks like Elasticsearch (Lucene) uses a tiered merge strategy of 10
+# segments of 512 megs per
+# https://lucene.apache.org/core/8_1_0/core/org/apache/lucene/index/TieredMergePolicy.html#setSegmentsPerTier(double)
```
gndcshv, line 56: ES specific defaults are set here: https://github.com/elastic/elasticsearch/blob/v6.8.2/server/src/main/java/org/elasticsearch/index/MergePolicyConfig.java
It is a tiered merge strategy with 10 segments per tier and max merged segment size is 5GB.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+# (FIXME) Verify what elastic merge actually does
+def _es_io_per_read(node_size_gib, segment_size_mb=512):
+    gb = node_size_gib * 1024
```
gndcshv, line 59: gb or mb?

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+    max_local_disk_gib: int = 4096,
+    max_regional_size: int = 240,
+    max_rps_to_disk: int = 500,
+) -> Optional[CapacityPlan]:
+
+    # Netflix Elasticsearch doesn't like to deploy on really small instances
+    if instance.cpu < 2 or instance.ram_gib < 14:
+        return None
+
+    # (FIXME): Need elasticsearch input
```
gndcshv, line 164: Instances with attached SSDs worked for us so far. No plans to support EBS.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+    else:
+        cluster.cluster_params = params
+
+
+# pylint: disable=too-many-locals
+def _estimate_elasticsearch_cluster_zonal(
+    instance: Instance,
+    drive: Drive,
+    desires: CapacityDesires,
+    zones_per_region: int = 3,
```
gndcshv, line 153: JFYI in us-west-1 there will 2 AZs.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+        needed_memory_gib=int(requirement.mem_gib.mid),
+        needed_network_mbps=requirement.network_mbps.mid,
+        # Assume that by provisioning enough memory we'll get
+        # a 90% hit rate, but take into account the reads per read
+        # from the per node dataset using leveled compaction
+        # FIXME: I feel like this can be improved
+        required_disk_ios=lambda x: _es_io_per_read(x) * math.ceil(0.1 * rps),
+        # Elasticsearch requires ephemeral disks to be % full because leveled
+        # compaction can make progress as long as there is some headroom
+        required_disk_space=lambda x: x * 1.25,
```
gndcshv, line 216: Existing ES self-service provisions clusters so that data instances have at least 30% of free space left.

service_capacity_modeling/models/org/netflix/elasticsearch.py
gndcshv: This is true for data nodes, master nodes run on EBS-based instances like r5.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+        needed_cores=int(requirement.cpu_cores.mid),
+        needed_disk_gib=int(requirement.disk_gib.mid),
+        needed_memory_gib=int(requirement.mem_gib.mid),
+        needed_network_mbps=requirement.network_mbps.mid,
+        # Assume that by provisioning enough memory we'll get
+        # a 90% hit rate, but take into account the reads per read
+        # from the per node dataset using leveled compaction
+        # FIXME: I feel like this can be improved
+        required_disk_ios=lambda x: _es_io_per_read(x) * math.ceil(0.1 * rps),
+        # Elasticsearch requires ephemeral disks to be % full because leveled
```
gndcshv, line 214: Leveled compaction?

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+        instance: Instance,
+        drive: Drive,
+        context: RegionContext,
+        desires: CapacityDesires,
+        extra_model_arguments: Dict[str, Any],
+    ) -> Optional[CapacityPlan]:
+        # (FIXME): Need elasticsearch input
+        # TODO: Use durability requirements to compute RF.
+        copies_per_region: int = _target_rf(
+            desires, extra_model_arguments.get("copies_per_region", None)
```
gndcshv, line 270: Existing ES self service assumes there will be 3 copies, for large use cases it makes sense to run with 2 copies to save on cluster cost. We can assume it's 3 by default and let users change it to 2 if they want.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+            (
+                "max_regional_size",
+                # Twice the size of our largest cluster
+                "int = 240",
+                "What is the maximum size of a cluster in this region",
+            ),
+            (
+                "max_local_disk_gib",
+                # Nodes larger than 4 TiB are painful to recover
+                "int = 4096",
```
gndcshv, line 310: I think it makes sense to raise this slightly to include `i3en.2xlarge` which have 5,000 GB, so set `max_local_disk_gib` to `5000` to be slightly over?

service_capacity_modeling/models/org/netflix/elasticsearch.py
jolynch: :+1: I think this IO estimate should be accurate then (although if there are no cloud drives involved it shouldn't really matter).

service_capacity_modeling/models/org/netflix/elasticsearch.py
jolynch: :shrug: changed to `size_mb`

service_capacity_modeling/models/org/netflix/elasticsearch.py
jolynch: yeah right now the model is only spitting out the data clusters, can add masters in and routing in a followup.

service_capacity_modeling/models/org/netflix/elasticsearch.py
jolynch: Ok changing to 1.4 (1/ 0.7)

service_capacity_modeling/models/org/netflix/elasticsearch.py
jolynch: tiered merging ;-) changing

service_capacity_modeling/models/org/netflix/elasticsearch.py
jolynch: this also assumes 3 by default, will only reduce if they reduce their durability requirement.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+    # Keep half of the cores free for background work (merging mostly)
+    needed_cores = math.ceil(sqrt_staffed_cores(desires) * 1.5)
+    # Keep half of the bandwidth available for backup
+    needed_network_mbps = simple_network_mbps(desires) * 2
+
+    needed_disk = math.ceil(
+        (1.0 / desires.data_shape.estimated_compression_ratio.mid)
+        * desires.data_shape.estimated_state_size_gib.mid
+        * copies_per_region,
+    )
```
szimmer1, line 89: Shouldn't needed_cores, needed_network_mbps, and needed_disk be divided by zones_per_region? It seems like these are regional calculations though Elasticsearch shards are distributed across AZ's

service_capacity_modeling/models/org/netflix/elasticsearch.py
gndcshv: Agree

service_capacity_modeling/models/org/netflix/elasticsearch.py
hgp2k: We have some clusters with i3en.3xl with 7500 GB disk space.

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -0,0 +1,471 @@
...
+
+    @staticmethod
+    def default_desires(user_desires, extra_model_arguments: Dict[str, Any]):
+        acceptable_consistency = set(
+            (
+                AccessConsistency.best_effort,
+                AccessConsistency.eventual,
+                # This will make writes very expensive due to refresh and
+                # waitfor
+                AccessConsistency.read_your_writes,
```
jolynch, line 328: Let's remove read_your_writes, seems aggressive refresh intervals is not something we want to encourage generally.

PR #13, opened 2021-06-29
Initial commit for TimeSeries

rajivshringi:
Model for netflix timeseries. Test verification pending.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/time_series.py
+++ b/service_capacity_modeling/models/org/netflix/time_series.py
@@ -1,0 +1,152 @@
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import Sequence
+from typing import Tuple
+
+from .stateless_java import nflx_java_app_capacity_model
+from service_capacity_modeling.interface import AccessPattern
+from service_capacity_modeling.interface import CapacityDesires
+from service_capacity_modeling.interface import CapacityPlan
+from service_capacity_modeling.interface import DataShape
+from service_capacity_modeling.interface import Drive
+from service_capacity_modeling.interface import FixedInterval
+from service_capacity_modeling.interface import Instance
+from service_capacity_modeling.interface import Interval
+from service_capacity_modeling.interface import QueryPattern
+from service_capacity_modeling.interface import RegionContext
+from service_capacity_modeling.models import CapacityModel
+
+
+class NflxTimeSeriesCapacityModel(CapacityModel):
+    @staticmethod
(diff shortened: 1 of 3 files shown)


service_capacity_modeling/models/org/netflix/time_series.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @staticmethod
+    def extra_model_arguments() -> Sequence[Tuple[str, str, str]]:
+        return nflx_java_app_capacity_model.extra_model_arguments()
+
+    @staticmethod
+    def compose_with(
+        user_desires: CapacityDesires, extra_model_arguments: Dict[str, Any]
+    ) -> Tuple[str, ...]:
+        # In the future depending on the user desire we might need EVCache
```
jolynch, line 59: nit: depending on the extra model arguments we'll need to add Elasticsearch

tests/netflix/test_time_series.py
```diff
@@ -0,0 +1,78 @@
+from service_capacity_modeling.capacity_planner import planner
+from service_capacity_modeling.interface import CapacityDesires
+from service_capacity_modeling.interface import DataShape
+from service_capacity_modeling.interface import Interval
+from service_capacity_modeling.interface import QueryPattern
+
+
+def test_timeseries_increasing_qps_simple():
+    qps_values = (100, 1000, 10_000, 100_000)
```
jolynch, line 9: Since this is a pretty slow test to run on every model we could just have 1k, 10k, 100k

jolynch (approved):

service_capacity_modeling/models/org/netflix/time_series.py
rajivshringi: How do you envision the arguments to be? Would that be an enum that ties to a model name or the model name itself? Should we add that in the next iteration when we are adding support for ES?

tests/netflix/test_time_series.py
rajivshringi: Updated

service_capacity_modeling/models/org/netflix/time_series.py
jolynch: The extra model arguments come from the free-form parameters. So I imagine that in the UI we'd have "Require best-effort cross event search?" and that would set a boolean flag "search=True" or some such in extra_model_arguments.

Then this function can return both elastic and C* when we are composing them.

jolynch (approved):

PR #15, opened 2021-07-12
Removes approver constraint for tier 0 zookeeper

szimmer1:


jolynch:
Isn't there a test that checks tier 0s have an approver? I think we have to remove the unit test now.

PR #16, opened 2021-07-21
adding a composite model for entity service.

abersnaze:


tests/netflix/test_entity.py
```diff
@@ -0,0 +1,106 @@
...
+            assert es_plan.instance.drive.size_gib >= 70
+
+        assert es_plan_family in ("m5d", "m5d.xlarge", "r5d", "r5d.large")
+        es_results_trend.append(
+            (
+                es_plan_cpu,
+                es_plan_cost,
+                cap_plan.least_regret[0].requirements.zonal[0],
+            )
+        )
```
abersnaze, line 77: I don't have the experience to know if these values are correct.

PR #27, opened 2021-11-03
EBS GP2 volumes should not start out larger than ~5.4 TiB

jolynch:
Should fix the issues with highly dense clusters.


jolynch:
So a couple of notable changes here.

1. We're limiting EBS attachement to 5.4 GiB where EBS maxes out its IO budget. As part of the gp3/io2 improvements we'll need to make this calculation a little more robust than 1/3 max size but can go from here.
2. Cassandra clusters will now scale up to 192 nodes per region if it is better that way.

Issue #20, opened 2021-09-01
Add support for gp3 volume types


jolynch:
Now that gp3 is a thing let's add it to the hardware descriptions so we can use it.

jolynch:
Both are supported in 'alpha' now.

PR #29, opened 2022-01-13
Add io2 drives and allow ES to use it if shown it

jolynch:
This starts working towards exploring the new gp3, io2 and io2 block express cloud drive families AWS has released. Makes progress towards #20 but since @jsjeannotte has requested we focus on io2 for elasticsearch.

Also adds the ability for callers to request specific hardware (instance families or drive families).


jolynch:
Will close #20

PR #35, opened 2022-03-08
moving the math to scale/offset methods

abersnaze:
compute the cassandra storage from the writes.


service_capacity_modeling/interface.py
```diff
@@ -74,6 +74,38 @@ def __hash__(self):
     def __eq__(self, other):
         return self.__hash__() == other.__hash__()
 
+    def scale(self, factor: float):
```
jolynch, line 77: nit: type annotation on return (use a string to avoid circular reference
```
def scale(self, factor: float) -> "Interval":
```

service_capacity_modeling/interface.py
```diff
@@ -74,6 +74,38 @@ def __hash__(self):
...
+            mid=self.mid * factor,
+            high=self.high * factor,
+            confidence=self.confidence,
+            model_with=self.model_with,
+            allow_simulate=self.allow_simulate,
+            minimum_value=minimum_value,
+            maximum_value=maximum_value,
+        )
+
+    def offset(self, delta):
```
jolynch, line 95: nit: type annotation
```
def scale(self, delta: float) -> "Interval":
```

jolynch (approved):
Awesome, fix (or not) and ship :-)

service_capacity_modeling/interface.py
abersnaze: i had it them in there at one point but was playing with the idea implementing them as operator overloads. It was way more code and figured you wouldn't like it and switched back.

PR #40, opened 2022-05-24
change the model to not serialize the defaults.

abersnaze:
i tip from Itay so we don't have to always remember to `.dict(exclude_unset=True)`


service_capacity_modeling/interface.py
```diff
@@ -13,6 +13,17 @@
 from pydantic import BaseModel
 from pydantic import Field
 
+
+class ExcludeUnsetModel(BaseModel):
+    def dict(self, *args, **kwargs):
+        kwargs["exclude_unset"] = True
```
jolynch, line 19: Can we check if the exclude_unset param was explicitly passed and is False, and if so don't exclude it?

jolynch (approved):
I am a little concerned, but not overly.

jolynch (approved):

PR #42, opened 2023-02-09
Force the planner to use a certain instance type

anujphadke:


service_capacity_modeling/capacity_planner.py
```diff
@@ -365,16 +374,17 @@ def _plan_certain(
 
         plans = []
         for instance in hardware.instances.values():
+            exact_match = instance.family_separator in instance.family
```
jolynch, line 377: This will always return true false. I think you want to do the for loop like I suggested inside the method against the strings in instance_families

service_capacity_modeling/capacity_planner.py
```diff
@@ -578,4 +588,4 @@ def _sub_models(
 
 
 planner = CapacityPlanner()
-planner.register_group(netflix.models)
+planner.register_group(netflix.models)
```
jolynch, line 591: nitpick: add back in a newline (UNIX mode not Windows)

jolynch (changes requested):

service_capacity_modeling/capacity_planner.py
anujphadke: @jolynch - Does it make sense to have separate functions? 
def _allow_instance
and
def _allow_drives

We use this function for the disk drives below on line 377. Should we keep this change specific to instance?
https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/main/service_capacity_modeling/capacity_planner.py#L377

jolynch:
I believe the allowed families, platforms and lifecycles give the appropriate filtering for this now https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/7f4d537a9167940ad7ee4b16ffacf176d7477d45/service_capacity_modeling/capacity_planner.py#L509C9-L509C26 

Closing for now we can re-open and re-tool if we need to.

PR #43, opened 2023-02-16
changes to surface networking costs

pkarumanchi9:

PR #44, opened 2023-02-22
Surface network transfer costs

jolynch:
Previously we didn't surface the costs of network transfer which might be quite significant for larger services like EVCache and Cassandra. Now models can include those cost components if they wish.

Original patch by @pkarumanchi9 in #43 . Reworked with him to move to network costs.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/evcache.py
+++ b/service_capacity_modeling/models/org/netflix/evcache.py
@@ -233,7 +233,7 @@ def _estimate_evcache_cluster_zonal(
         modified = desires.copy(deep=True)
         # Assume that DELETES replicating cross region mean 128 bytes
         # of key per evict.
-        modified.query_pattern.estimated_mean_write_size_bytes.mid = 128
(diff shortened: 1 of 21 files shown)


service_capacity_modeling/models/org/netflix/evcache.py
```diff
@@ -214,13 +216,20 @@ def reserve_memory(instance_mem_gib):
     if cluster.count > (max_regional_size // zones_per_region):
         return None
 
+    services = network_services("evcache", context, desires, copies_per_region)
     ec2_cost = zones_per_region * cluster.annual_cost
 
+    # Account for the clusters and replication costs
+    evcache_costs = {"evcache.zonal-clusters": ec2_cost}
+    for s in services:
+        evcache_costs[f"{s.service_type}"] = s.annual_cost
```
pkarumanchi9, line 225: cross region costs are not applicable by default. Might need to check on a flag before giving the estimates.

jolynch: :+1: Done

PR #46, opened 2023-02-27
Fix off-by-N error on intra costs

jolynch:
Previously we did not account for the number of regions when billing cross zone costs. Now we do by multiplying the intra-region (inter-zone) costs by the number of regions. Note that non global databases simply will not add these services to their cost model.

cc @pkarumanchi9


abersnaze (approved):
LGTM

PR #47, opened 2023-03-13
Add 6th Generation Hardware

jolynch:
Adds the i4i, m6id, and m6i families which are now generally available. Also refactors our test suite to be less fragile and applies pre-commit across the entire codebase.


jolynch:
@arunagrawal84 once this lands internally you should see i4is recommended.

PR #48, opened 2023-03-17
Add EVCache composition in KV

alexsyeo:


service_capacity_modeling/models/org/netflix/key_value.py
```diff
@@ -60,9 +59,30 @@ def extra_model_arguments_schema() -> Dict[str, Any]:
...
+                       (user_desires.query_pattern.estimated_read_per_second.mid > 100000 and
+                        (user_desires.query_pattern.estimated_read_per_second.mid / user_desires.query_pattern.estimated_write_per_second.mid > 0.9)))
+
+        if use_evcache:
+            def _modify_evcache_desires(
+                desires: CapacityDesires,
+            ) -> CapacityDesires:
+                relaxed = desires.copy(deep=True)
+
+                # TODO: Confirm if any other desire properties should be modified for evcache
```
szimmer1, line 76: Have we discussed target working set size (ie. data size in the cache)? I don't think we provision evcache for the entire dataset; in the absence of more workload info (latency target, request cardinality, etc) we could just set a heuristic of 10% data set size to limit overprovisioning

alexsyeo: Just pushed a change for scaling down the data to 10%. I wasn't sure if this is what you meant or setting the "estimated_working_set_percent" property to 10%

jolynch (commented):
I'm working on fixing the mypy breakage in the i4i branch, should have that in here shortly for you to rebase

service_capacity_modeling/models/org/netflix/key_value.py
szimmer1: Ok i was not aware of the working set percent property. It looks like we set it in a number of our [shards](https://sourcegraph.netflix.net/search?q=context:global+estimated_working_set_percent+&patternType=standard&sm=1&groupBy=repo).

It looks like the evcache model [understands this property](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/dd6392ef45c4cc52f01cadbebb7b22375f566208/service_capacity_modeling/models/common.py#L392) the way i explained above (ie. the estimated % of total data that should be served from cache).

Probably the right thing to do here is pass the property through to the evcache model, which you were already doing. So let's revert the change to scale down data size.

szimmer1: cc @jolynch

szimmer1 (approved):

szimmer1 (approved):

jolynch (approved):

PR #53, opened 2023-05-16
Feature/shengweiw/aurora hardware

ShengweiWang:
1. Using db.xx instances instead of normal ec2 instances, to get a more accurate cost estimation
2. Adding IO cost estimation

### DIFF
--- a/service_capacity_modeling/models/org/netflix/aurora.py
+++ b/service_capacity_modeling/models/org/netflix/aurora.py
@@ -91,6 +91,24 @@ def _rds_required_disk_ios(disk_size_gib: int, db_type: str, btree_fan_out: int 
     return math.log(pages, btree_fan_out)
 
 
+# This is a start, we should iterate based on the actual work load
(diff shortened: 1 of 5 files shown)


service_capacity_modeling/models/org/netflix/aurora.py
```diff
@@ -161,7 +180,7 @@ def _estimate_aurora_regional(
     extra_model_arguments: Dict[str, Any],
 ) -> Optional[CapacityPlan]:
     instance_family = instance.family
-    if instance_family not in ("x2g", "r6g", "r6i", "r5", "t4g"):  # TODO: split db instance and ec2 instance
+    if instance_family not in ("db.x2g", "db.r6g", "db.r6i", "db.r5", "db.t4g"):
```
abersnaze, line 183: could this be `not instance_family.startswith('db.')`?

ShengweiWang: There are only some of the instance families are supported in Aurora. For example, db.m5 is supported by RDS but not Aurora.

PR #54, opened 2023-05-25
Adding platform info for db instances

ShengweiWang:
1. Add platform info for db instances
2. Bug fixing in instance interface to support .deprecated instances
3. use postgres as the default Aurora engine

### DIFF
--- a/service_capacity_modeling/hardware/profiles/shapes/aws.json
+++ b/service_capacity_modeling/hardware/profiles/shapes/aws.json
@@ -726,7 +726,7 @@
       "ram_gib": 15.71,
       "net_mbps": 500,
       "drive": null,
-      "platforms": "Aurora MySQL,Aurora PostgreSQL"
+      "platforms": ["Aurora MySQL","Aurora PostgreSQL"]
     },
     "db.r5.xlarge": {
       "name": "db.r5.xlarge",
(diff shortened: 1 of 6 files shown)


service_capacity_modeling/hardware/profiles/shapes/aws.json
```diff
@@ -725,63 +725,71 @@
       "cpu_ghz": 3.1,
       "ram_gib": 15.71,
       "net_mbps": 500,
-      "drive": null
+      "drive": null,
+      "platforms": "Aurora MySQL,Aurora PostgreSQL"
```
jolynch, line 729: Can you just make this a list instead of a string, e.g. ["aurora-mysql", "aurora-postgresql"] ?

service_capacity_modeling/interface.py
```diff
@@ -271,16 +271,21 @@ class Instance(ExcludeUnsetModel):
     drive: Optional[Drive]
     annual_cost: float = 0
     lifecycle: Lifecycle = Lifecycle.stable
+    platforms: str = "EC2"
```
jolynch, line 274: platforms: List[str] = ["x86-intel"]

jolynch (commented):
Doesn't this need changes to the plan logic to avoid unsupported platforms for a model (and the model has a method of "supported platforms" or such)? Perhaps something in capacity_planner.py?

service_capacity_modeling/hardware/profiles/shapes/aws.json
jolynch: Did in a followup.

PR #55, opened 2023-06-09
Allow customized minimal vcpu for crdb

ShengweiWang:


jolynch (approved):

service_capacity_modeling/models/org/netflix/crdb.py
```diff
@@ -138,10 +138,10 @@ def _estimate_cockroachdb_cluster_zonal(
     max_local_disk_gib: int = 2048,
     max_regional_size: int = 288,
     max_rps_to_disk: int = 500,
+    min_vcpu_per_instance: int = 4,
```
jolynch, line 141: minor nit: I'd just call it `cpu` since vcpu isn't really a thing outside AWS

ShengweiWang: https://www.cockroachlabs.com/docs/stable/common-issues-to-monitor.html#cpu-planning CRDB uses this term broadly, more like a cloud thing?

PR #61, opened 2023-07-06
CRDB working set tests

jolynch:
Just to add a unit test of the current SLOs we are offering for CRDB and the neccesary working set.


jolynch:
I don't think we need this anymore.

PR #62, opened 2023-07-10
Pydantic V2 ensure compatibility

jolynch:
Fixed up compatibility with `pydantic` V2, mostly stricter validations on data (e.g. ints can't be floats, providing superclass instead of subclass, etc ...).

Note that there are a _ton_ of deprecation warnings in V2 since we still use V1 APIs like `__fields__`, but most of the replacement methods aren't available in V1 so we can't use those either. I resolved the issues that were throwing active type exceptions, but left the use of V1 apis for now since that works in both V1 and V2 at least. The internal app that includes this library is pinned to <V2 for now, once that upgrades we can upgrade this to use v2 and just hope nobody uses this library inside a V1 app (which would promptly explode).

Also adds to the matrix to run the unit test suite against both v1 and v2 pydantic and adds two more python versions while I'm there.

### DIFF
--- a/.github/workflows/python-build.yml
+++ b/.github/workflows/python-build.yml
@@ -18,5 +18,5 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install Tox and any other packages
       run: pip install tox
-    - name: Run Tox
+    - name: Run Tox (pydantic v1)
       run: tox -e py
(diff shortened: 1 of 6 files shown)


jolynch:
I think that the build is timing out because of the number of warnings being emitted slows down the test run, trying a push without V2 warning all the time.

jolynch:
Pinning to <v2 fixed the build, I think it's the warnings being logged :-(

PR #63, opened 2023-07-13
Tweaks evcache model and adds experiment

akashdeepgoel:


service_capacity_modeling/capacity_planner.py
```diff
@@ -377,19 +377,30 @@ def _plan_certain(
             if per_instance_mem > instance.ram_gib:
                 continue
 
-            for drive in hardware.drives.values():
-                if not _allow_hardware(drive.name, drive.lifecycle, drives, lifecycles):
-                    continue
-
+            if "evcache" in model_name:
```
jolynch, line 380: This is the generic planner interface, it's the wrong level of abstraction to add model names. The model itself will be shown all possible instance+drive combinations, it can ignore ones it doesn't want.

service_capacity_modeling/models/org/netflix/evcache.py
```diff
@@ -261,9 +262,11 @@ def reserve_memory(instance_mem_gib):
         )
 
     ec2_cost = zones_per_region * cluster.annual_cost
+    spread_cost = 100000.0/float(cluster.count)
```
jolynch, line 265: I'd make it proportional to the existing cost rather than a fixed cost. So e.g. take the ec2 cost and add a decreasing tax by node count. Example
```
# g = number of nodes
tax = ec2_cost * g(x)
def tax(x, max_tax=0.10, min_tax=0.0):
    if x < 2:
        return max_tax
    elif x >= 32:
        return min_tax
    else:
        return min_tax + (max_tax - x * (max_tax - min_tax) / 30)
```

jolynch (changes requested):

service_capacity_modeling/models/org/netflix/evcache.py
```diff
@@ -154,6 +152,9 @@ def _estimate_evcache_cluster_zonal(
     cross_region_replication: Replication = Replication.none,
 ) -> Optional[CapacityPlan]:
 
+    if "m6" in instance.family or "r6" in instance.family:
```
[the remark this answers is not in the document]
pkarumanchi9: Makes sense. We shouldn't be taking into consideration things like reserved capacity, so it makes to delete these 2 lines altogether.

jolynch (approved):
Awesome thank you @akashdeepgoel

PR #65, opened 2023-08-30
Tune memory on kafka clusters

jolynch:
Keep the last 10 minutes of data hot in RAM for consumers, and don't worry about the rest

### DIFF
--- a/service_capacity_modeling/interface.py
+++ b/service_capacity_modeling/interface.py
@@ -19,6 +19,8 @@ from pydantic import Field
 
 
 GIB_IN_BYTES = 1024 * 1024 * 1024
+MIB_IN_BYTES = 1024 * 1024
+MEGABIT_IN_BYTES = (1000 * 1000) / 8
 
 
 class ExcludeUnsetModel(BaseModel):
@@ -207,6 +209,10 @@ class Drive(ExcludeUnsetModel):
     max_scale_io_per_s: int = 0
     # How large is an "IO" against this device
     block_size_kib: int = 4
+    # When sequential how much IO is grouped into a single "IO"
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-io-characteristics.html
+    # Some cloud drives can group sequential ops together and DBs take advantage
+    group_size_kib: int = 4
 
     lifecycle: Lifecycle = Lifecycle.stable
     compatible_families: List[str] = []
@@ -225,6 +231,14 @@ class Drive(ExcludeUnsetModel):
     write_io_latency_ms: FixedInterval = FixedInterval(
         low=0.6, mid=2, high=3, confidence=0.9
     )
+
+    @property
+    def rand_io_size_kib(self) -> int:
+        return self.block_size_kib
+
+    @property
+    def seq_io_size_kib(self) -> int:
+        return max(self.block_size_kib, self.group_size_kib)
 
     @property
     def max_size_gib(self):
--- a/service_capacity_modeling/models/org/netflix/kafka.py
+++ b/service_capacity_modeling/models/org/netflix/kafka.py
@@ -24,15 +24,15 @@ from service_capacity_modeling.interface import GIB_IN_BYTES
 from service_capacity_modeling.interface import GlobalConsistency
 from service_capacity_modeling.interface import Instance
 from service_capacity_modeling.interface import Interval
+from service_capacity_modeling.interface import MEGABIT_IN_BYTES
+from service_capacity_modeling.interface import MIB_IN_BYTES
 from service_capacity_modeling.interface import QueryPattern
 from service_capacity_modeling.interface import RegionContext
 from service_capacity_modeling.interface import Requirements
 from service_capacity_modeling.models import CapacityModel
 from service_capacity_modeling.models.common import compute_stateful_zone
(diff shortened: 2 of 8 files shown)


service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -207,7 +210,8 @@ def _estimate_kafka_cluster_zonal(
         cluster_size=lambda x: next_n(x, zones_per_region),
         min_count=max(min_count, 1),
         # Sidecars and Variable OS Memory
-        reserve_memory=lambda x: base_mem,
+        # Kafka currently uses 8GiB fixed, might want to change to min(30, x // 2)
```
nickmahilani, line 213: Thinking more about this, we do not want to allocate a lot of RAM to JVM heap for Kafka as we want to leave most memory for the page cache so consumers don't have to do disk reads. The JVM heap usage is low in kafka brokers even at high scale due to zero-copy transfers

nickmahilani (approved):
Thanks for the quick PR ! LGTM

service_capacity_modeling/models/org/netflix/kafka.py
jolynch: The heap is probably pretty useful for Kafka, but agreed if y'all end up changing your heap formula we should refactor this.

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -76,14 +78,27 @@ def _estimate_kafka_requirement(
     needed_cores = sqrt_staffed_cores(normalized_to_mib)
 
     # (Nick): Keep 40% of available bandwidth for node recovery
-    needed_network_mbps = simple_network_mbps(desires) * 1.40
+    # (Joey): For kafka BW = BW_write + BW_reads
+    #   let X = input write
+    #   BW_write = [BW_copies] + BW_leader
+    #   BW_write = [2X * (RF - 1)] + X
+    #   BW_write = 2X * RF - 2X + X = 2X * RF - X
+    #   BW_write = X * (2RF - 1)
```
nickmahilani, line 86: I am trying to understand why scale up by `2RF - 1` here. This would mean `BW_write = X*3` for replication factor of 2

I was expecting `BW_write = X * RF` here as the total write BW expected for the cluster

jolynch: Yeah this math assumes that the first write goes to the right replica. In that case you have X `in` to the first replica, and X `out` + X `in` per replica.

If kafka doesn't route the write to the correct first replica (in C* land we call that "token aware routing"), then the BW is X `in` to the coordinator + `RF * 2 * X` for each replica.

nickmahilani: I see, thanks for sharing the C* context. For Kafka, the kafka client would always directly write to the right leader replica for all writes, so `Bwrite = X * RF`

service_capacity_modeling/interface.py
```diff
@@ -207,6 +209,10 @@ class Drive(ExcludeUnsetModel):
     max_scale_io_per_s: int = 0
     # How large is an "IO" against this device
     block_size_kib: int = 4
+    # When sequential how much IO is grouped into a single "IO"
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-io-characteristics.html
+    # Some cloud drives can group sequential ops together and DBs take advantage
+    group_size_kib: int = 4
```
nickmahilani, line 215: awesome ! thanks for adopting the model to accommodate the seq IO EBS

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -120,20 +137,22 @@ def _upsert_params(cluster, params):
...
+    # Get enough disk read IO capacity for some reads
+    # In practice we have cache reducing this by 99% or more
+    read_ios = rps * 0.05
     # Recover the node in 60 minutes, to do that we need
     size_kib = size_gib * (1024 * 1024)
-    recovery_ios = max(1, size_kib / block_size_kib)
-    return read_ios + recovery_ios
+    recovery_ios = max(1, size_kib / io_size_kib) / recovery_seconds
+    # Leave 50% headroom for read IOs since generally we will hit cache
+    return (read_ios + int(round(recovery_ios))) * 1.5
```
nickmahilani, line 148: nice !

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -367,12 +412,12 @@ def default_desires(user_desires, extra_model_arguments: Dict[str, Any]):
                 # This = 1000 / 14 = 71 millisecond / MiB
                 # To read 1 MiB costs 71 millisecond of CPU time
                 estimated_mean_read_latency_ms=Interval(
-                    low=50, mid=70, high=100, confidence=0.98
+                    low=30, mid=40, high=100, confidence=0.98
```
nickmahilani, line 415: thinking we can probably lower this to `low=20, mid=30, high=75` based on empirical data for both reads and writes.

service_capacity_modeling/models/org/netflix/kafka.py
jolynch: Fixed per your suggestions

PR #69, opened 2023-09-29
Introduce Dynamo model

raj14243:


service_capacity_modeling/models/common.py
```diff
@@ -129,7 +129,7 @@ def network_services(
     # inter region. This is the number of regions minus 1
     inter_txfer = context.services.get("net.inter.region", None)
     if inter_txfer:
-        price_per_gib = inter_txfer.annual_cost_per_gib
+        price_per_gib = inter_txfer.annual_cost_per_gib[0][1]
```
jolynch, line 132: If we do the annual_cost then we'd just size the net.inter.region `Service` and then call annual_cost and multiply by number of regions

service_capacity_modeling/models/org/netflix/__init__.py
```diff
@@ -34,4 +35,5 @@ def models():
         "org.netflix.aurora": nflx_aurora_capacity_model,
         "org.netflix.postgres": nflx_postgres_capacity_model,
         "org.netflix.kafka": nflx_kafka_capacity_model,
+        "org.netflix.ddb": nflx_ddb_capacity_model,
```
jolynch, line 38: nit: can we spell this out to match the others (I know rds didn't but that was a miss). Perhaps `dynamodb`

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+from service_capacity_modeling.interface import QueryPattern
+from service_capacity_modeling.interface import RegionContext
+from service_capacity_modeling.interface import Requirements
+from service_capacity_modeling.interface import Service
+from service_capacity_modeling.interface import ServiceCapacity
+from service_capacity_modeling.models import CapacityModel
+
+
+class NflxDynamoDBArguments(BaseModel):
+    number_of_regions: int = Field(
```
jolynch, line 34: This should be available in the RegionContext in structured arguments, no need for a model specific argument for this.

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+        "then all-or-nothing guarantee for writes. "
+        "Default is zero, assuming most writes are single item puts",
+    )
+    transactional_read_percent: float = Field(
+        default=0,
+        description="Coordinated reads, if multiple items in "
+        "single read then all-or-nothing guarantee for read. "
+        "Default is zero",
+    )
+    estimated_mean_item_size_bytes: int = Field(
```
jolynch, line 52: I'm a little confused why we need this argument when we have the write size argument but :shrug:

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+        default=0,
+        description="Estimated avg item size. If not supplies, "
+        "then estimated_mean_write_size_bytes from desires will "
+        "be used",
+    )
+
+
+class _ReadPlan(BaseModel):
```
jolynch, line 60: If you're returning these from the model as json anywhere you probably want to inherit from ExcludeUnsetModel

tests/netflix/test_ddb.py
```diff
@@ -0,0 +1,314 @@
...
+    assert mean_plan[0].requirements.regional[0].context == regional_req
+    annual_costs = {
+        "dynamo.regional-writes": Decimal("0.0"),
+        "dynamo.regional-reads": Decimal("0.0"),
+        "dynamo.regional-storage": Decimal("300.0"),
+        "dynamo.regional-transfer": Decimal("0.0"),
+        "dynamo.data-backup": Decimal("240.0"),
+    }
+    assert mean_plan[0].candidate_clusters.annual_costs == annual_costs
+    assert round(mean_plan[0].candidate_clusters.total_annual_cost, 0) == round(540, 0)
```
jolynch, line 83: If we're going to make exact assertions like this you probably want to use plan_certain instead of plan since the latter still will simulate the inputs.

Also I think range assertions are probably more robust longer term (e.g. in this case >100 and < 1000).

Same advice for the rest of the file, `plan` is non deterministic (but it is reproducible) within a range of valid outputs. `plan_certain` has no simulation.

service_capacity_modeling/interface.py
```diff
@@ -330,6 +334,12 @@ def family(self):
...
         return self.name.rsplit(self.family_separator, 1)[1]
 
+    @staticmethod
+    def get_managed_instance() -> Instance:
+        return Instance(
+            name="managed.0", cpu=0, cpu_ghz=0, ram_gib=0, net_mbps=0, drive=None
+        )
+
 
 class Service(ExcludeUnsetModel):
```
jolynch, line 344: Just leaving a note that when we apply our pricing in antigravity we need to update the type of the annual_cost number.

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+    transactional_write_percent: float = extra_model_arguments.get(
+        "transactional_write_percent", 0
+    )
+    assert transactional_write_percent >= 0.0
+    baseline_wcus_non_transactional = (
+        desires.query_pattern.estimated_write_per_second.mid
+        * (1 - transactional_write_percent)
+        * rounded_wcus_per_item
+    )
+    baseline_wcus_transactional = (
```
jolynch, line 187: nit: If you have them handle breadcrumbs to AWS docs that explain these various factors (e.g. the 2 wcu per item etc) might be useful.

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+        * rounded_wcus_per_item
+    )
+    # we do not break down reserved vs non-reserved
+    total_baseline_wcus = math.ceil(
+        baseline_wcus_non_transactional + baseline_wcus_transactional
+    )
+
+    # 8760 hours in a year
+    total_baseline_wcus_hours = 8760 * total_baseline_wcus
+    number_of_regions = _get_num_regions(extra_model_arguments)
```
jolynch, line 200: the RegionContext passed to this method has how many regions to plan for

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+    dynamo_service_standard = _get_dynamo_standard(context, extra_model_arguments)
+
+    # for global replication each region will additionally consume rWCUs
+    total_annual_cost_wcu = round(
+        (
+            number_of_regions
```
jolynch, line 206: Are rWCUs priced at the same thing as local WCU? Otherwise we want num_regions - 1 right? From https://aws.amazon.com/dynamodb/pricing/provisioned/ I think global writes are more expensive than local writes (0.00065 vs 0.000975 if I'm reading that right)

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+            * total_baseline_wcus_hours
+            * dynamo_service_standard.annual_cost_per_write_io
+        ),
+        2,
+    )
+    if number_of_regions > 1:
```
jolynch, line 212: I'm confused by this, wouldn't we always have local write costs and then if num_regions > 1 have replicated write units as well?

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+
+    @staticmethod
+    def default_desires(user_desires, extra_model_arguments: Dict[str, Any]):
+        acceptable_consistency = {
+            "same_region": {
+                None,
+                AccessConsistency.best_effort,
+                AccessConsistency.eventual,
+                AccessConsistency.read_your_writes,
+                AccessConsistency.never,
```
jolynch, line 468: I believe DDB can do up to `serializable` (essentially all consistency levels) in the same region

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+    extra_model_arguments: Dict[str, Any],
+) -> _DataTransferPlan:
+    number_of_regions = _get_num_regions(extra_model_arguments)
+    if not number_of_regions > 1:
+        return _DataTransferPlan(
+            total_data_transfer_gib=0, total_annual_data_transfer_cost=0
+        )
+    mean_item_size_bytes = _mean_item_size_bytes(desires, extra_model_arguments)
+    writes_per_second = desires.query_pattern.estimated_write_per_second.mid
+    # 31,536,000 seconds in a year
```
jolynch, line 320: nit: I often like to define constants at the top like this to make these multiplications easier to follow and check the units on

```
# Approximate seconds in a year
SECONDS_IN_YEAR= (365 * 24 * 60 * 60) # 31536000
# True time in a year if we want to be pedantic https://en.wikipedia.org/wiki/Year
SECONDS_IN_YEAR= 31556926

# 1GiB in bytes
GIB_IN_BYTES = 1024 * 1024 * 1024
```

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+                    write_plan.total_annual_write_cost
+                    + read_plan.total_annual_read_cost
+                    + storage_plan.total_annual_data_storage_cost
+                    + data_transfer_plan.total_annual_data_transfer_cost
+                ),
+                service_params={
+                    "read_capacity_units": read_plan.read_capacity_units,
+                    "write_capacity_units": (
+                        write_plan.write_capacity_units
+                        + write_plan.replicated_write_capacity_units
```
jolynch, line 415: Isn't there a storage as well?

jolynch (commented):
Awesome! Love the idea of the "managed" instance/drive to skip simulation.

Just a few pieces of feedback and some nitpicks.

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: This is not always available in the RegionContext, specifically if the approach is to plan for one region and then externally multiply with number of regions.

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: There are currently two arguments `estimated_mean_read_size_bytes` and `estimated_mean_write_size_bytes` and I wasn't sure which represents the item_size. So to disambiguate, introduced as extra_model_argument.

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: as soon as number of regions > 1 there is no WCUs everything is rWCUs even in the local

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: for creating ddb table, storage is not one of the inputs

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: let me update this

service_capacity_modeling/models/org/netflix/ddb.py
jolynch: It should always be accurate, if it is not that's a bug in the caller of this library. Even when planning in a single and extrapolating the AG endpoint passes the correct num_regions to the top level `plan` method  which fills in [num_regions](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/1c408dee7c2b3b4377e60f5137a86b6e8f347af7/service_capacity_modeling/capacity_planner.py#L455). The only thing that plan in a single region should do is use pricing in only one region.

service_capacity_modeling/models/org/netflix/ddb.py
jolynch: mean_write_size_bytes is the item size.

Reads is separate because most read apis can return _multiple_ items, so reads tend to be larger than writes.

service_capacity_modeling/models/org/netflix/ddb.py
jolynch: Oh really! Fascinating

service_capacity_modeling/models/org/netflix/ddb.py
jolynch: Yeah but that context is useful to have in the caller (what context was originally for heh)?

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: Lets fix the caller then

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: CapacityRequirement(
            requirement_type="dynamo-regional",
            core_reference_ghz=0,
            cpu_cores=certain_int(0),
            disk_gib=certain_float(storage_plan.total_data_storage_gib),
            context=requirement_context,
        )

service_capacity_modeling/models/org/netflix/ddb.py
```diff
@@ -0,0 +1,584 @@
...
+        dynamo_services = [
+            ServiceCapacity(
+                service_type="dynamo.standard",
+                annual_cost=(
+                    write_plan.total_annual_write_cost
+                    + read_plan.total_annual_read_cost
+                    + storage_plan.total_annual_data_storage_cost
+                    + data_transfer_plan.total_annual_data_transfer_cost
+                ),
+                service_params={
```
jolynch, line 411: Do you want to provide an autoscaling target based on scale and an upper bound based on cost (maybe an extra param for target upper bound cost and then allow autoscaling to that plus 2x above it)?

service_capacity_modeling/models/org/netflix/ddb.py
raj14243: These are not returned as json only used internally to pass b/w internal methods.

service_capacity_modeling/capacity_planner.py
```diff
@@ -325,6 +325,92 @@ def models(self) -> Dict[str, CapacityModel]:
     def hardware_shapes(self) -> HardwareShapes:
         return self._shapes
 
+    def _plan_percentiles(
```
raj14243, line 328: Introduced this method to address the following issue.

Previously the steps to generate percentile plan was
* generate percentile desires
* plan_certain using the generated desires 
Issue in the above is, when percentile desires are generated the attributes that are not supplied by user are replaced with defaults declared in CapacityDesires interface. This results in `parent_desires.merge_with` not merging the defaults i.e. not merging the model defaults for attributes that are not supplied by user.

To mitigate the above issue, we first generate desires by merging with model defaults and then generate the percentile desires using the merged desires.

jolynch: Nice find

jolynch (approved):

PR #74, opened 2023-10-13
Add cpu utilization as a parameter with which desires can be obtained

ramsrivatsa:
Use CPU utilization directly within extra model arguments. If we have utilization numbers, we can directly leverage them to obtain desired instance types. If we don't, then use the default reads and writes to compute CPU util which in turn can be used to obtain desired instance types. 

Also added pytest in the setup so that it can integrate seamlessly with pycharm.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -24,6 +24,7 @@ from service_capacity_modeling.interface import CapacityRequirement
 from service_capacity_modeling.interface import certain_float
 from service_capacity_modeling.interface import DataShape
 from service_capacity_modeling.interface import Drive
+from service_capacity_modeling.interface import Hardware
 from service_capacity_modeling.interface import Instance
 from service_capacity_modeling.interface import Interval
 from service_capacity_modeling.interface import interval
@@ -180,6 +181,21 @@ def model_desires_percentiles(
     d.data_shape = mean_ds
 
     return results, d
+
+
+def _set_instance_objects(
+    desires: CapacityDesires,
+    hardware: Hardware,
+):
+    if desires.current_clusters:
+        for zonal_cluster_capacity in desires.current_clusters.zonal:
+            zonal_cluster_capacity.cluster_instance = hardware.instances[
+                zonal_cluster_capacity.cluster_instance_name
+            ]
+        for regional_cluster_capacity in desires.current_clusters.regional:
+            regional_cluster_capacity.cluster_instance = hardware.instances[
+                regional_cluster_capacity.cluster_instance_name
+            ]
 
 
 def _allow_instance(
@@ -575,14 +591,9 @@ class CapacityPlanner:
         if len(allowed_drives) == 0:
             allowed_drives.update(hardware.drives.keys())
 
-        # Get current instance object if exists
-        if desires.current_clusters:
-            for zonal_cluster_capacity in desires.current_clusters.zonal:
-                zonal_cluster_capacity.cluster_instance = hardware.instances[zonal_cluster_capacity.cluster_instance_name]
-            for regional_cluster_capacity in desires.current_clusters.regional:
-                regional_cluster_capacity.cluster_instance = hardware.instances[regional_cluster_capacity.cluster_instance_name]
+        # Set current instance object if exists
+        _set_instance_objects(desires, hardware)
 
-        plans = []
         if model.run_hardware_simulation():
             for instance in hardware.instances.values():
                 if not _allow_instance(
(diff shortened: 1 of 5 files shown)


service_capacity_modeling/capacity_planner.py
```diff
@@ -503,6 +503,13 @@ def _plan_certain(
         if len(allowed_drives) == 0:
             allowed_drives.update(hardware.drives.keys())
 
+        # Get current instance type if exists
+        current_instance_name: str = extra_model_arguments.get("current_instance_name", None)
```
jolynch, line 507: let's not do this in extra args as it is generic. I'd put it into https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/main/service_capacity_modeling/interface.py#L623 as a [Clusters](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/f0dde0d182abac85febc6cd05e5438daa59e26eb/service_capacity_modeling/interface.py#L740) object. For example

```
class Desires()
...
    CapacityPlan existing_plan
```

Then in CapacityPlan you can have the existing [requirements](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/f0dde0d182abac85febc6cd05e5438daa59e26eb/service_capacity_modeling/interface.py#L723) expressed as cpu/mem/network.

You should be able to write a generic unit test where we give it a "wrong" cluster and then it corrects it to the "right" one.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -64,20 +64,26 @@ def _write_buffer_gib_zone(
 
 def _estimate_cassandra_requirement(
     instance: Instance,
+    current_instance: Instance,
```
jolynch, line 67: This is where I'd pass the existing plan and then use that for e.g. current sizing

jolynch (changes requested):

service_capacity_modeling/capacity_planner.py
ramsrivatsa: I did pass in the existing instance family in the interface. 

My current test case is overprovisioned (overscaled for the number of vcpus) and I've asserted it for the right one.

abersnaze:
i noticed a lot of whitespace change. @jolynch what is there a formatter that you used?

abersnaze:
I don't think the model should be aware what the current instance is.

I would like to see the current instance to be used as signal to the planner that one of the least_regret plans must be of that instance. that way we could:
- compare what the planner thinks should deploy vs what is.
- compare replans vs alternatives.

this also introduces more challenges to compositions of models since a kv cluster would likely have a different current instance than the composed cassandra.  I'm thinking we should depreciate composition and have the client invoke the planner for each model separately.

service_capacity_modeling/interface.py
```diff
@@ -634,6 +640,9 @@ class CapacityDesires(ExcludeUnsetModel):
     # What will the state look like
     data_shape: DataShape = DataShape()
 
+    # What is the current microarchitectural/system configuration of the system
+    current_capacity: CurrentCapacity = CurrentCapacity()
```
jolynch, line 644: This should probably be Optional[CurrentCapacity] = None. That way you don't have to have default values.

setup.py
```diff
@@ -14,6 +14,7 @@
         "numpy",
         'importlib_resources; python_version < "3.7"',
         "isodate",
+        "pytest",
```
jolynch, line 17: pytest should not be a library dependency.

jolynch (changes requested):

service_capacity_modeling/interface.py
```diff
@@ -620,6 +619,13 @@ class DataShape(ExcludeUnsetModel):
     )
 
 
+class CurrentCapacity(ExcludeUnsetModel):
+    current_instance_type: str = ""
```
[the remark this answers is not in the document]
jolynch: You could also mirror the output format and support multiple zonal and regional clusters, which would make this compatible with the existing output format and multiple input clusters

```
# Mirrors ClusterCapacity
class CurrentClusterCapacity:
  ... the fields you have

# Mirrors Requriements
class CurrentClusters:
  zonal: Sequence[CurrentClusterCapacity]
  regional: Sequence[CurrentClusterCapacity]
  services: Sequence[ServiceCapacity]  
```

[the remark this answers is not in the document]
abersnaze: count/size should be an interval too for cluster that can auto scale.

setup.py
ramsrivatsa: This helped us integrate well with pycharm IDE. Will be happy to remove if you want.

abersnaze (approved):

PR #78, opened 2023-10-24
add some kafka model constraint args

nickmahilani:


jolynch:
Test fails but patch lg2m

PR #82, opened 2023-10-30
Fixed cassandra capacity planner to normalize cpu utilization 

ramsrivatsa:
Fixed Cassandra capacity planner to normalize CPU utilization to reference clock. This is consistent with the rest of the capacity planning logic that assumes that the clock speed for which calculations are being pursued is the reference clock ghz.


jolynch:
We can follow up with factoring this out into a generic method so we don't have to repeat this complexity in every model that wants to rightsize.

PR #86, opened 2024-02-15
Decouple read size from write size in DDB model

william-schor:


jolynch:
@william-schor this seems like a reasonable idea to me, should we merge it?

raj14243:
> @william-schor this seems like a reasonable idea to me, should we merge it?

Yes, can be merged. This missed the radar

PR #88, opened 2024-03-21
Remove linting error for deprecated warning

ramsrivatsa:
Remove linting error for deprecated warning


jolynch:
Everything about https://discuss.python.org/t/deprecating-importlib-resources-legacy-api/11386/14?page=2 makes me sad.

Why can't they just keep path. There doesn't seem to me to be a strong reason to deprecate it

PR #91, opened 2024-05-27
Adds support for disk_io in evcache

akashdeepgoel:


pkarumanchi9 (changes requested):
let's add a test or two to validate this?

service_capacity_modeling/models/common.py
```diff
@@ -269,6 +272,17 @@ def compute_stateful_zone(
     ):
         disk_per_node = min(max_local_disk_gib, instance.drive.size_gib)
         count = max(count, math.ceil(needed_disk_gib / disk_per_node))
+        if adjusted_disk_io_needed != 0.0:
```
pkarumanchi9, line 275: for additional use cases like
cache warming, region squeeze and background activities like compaction, we can consider leaving additional 40% headroom

PR #95, opened 2024-10-30
Feature/remove ebs

ayushisingh29:


ramsrivatsa (approved):
Looks good

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -215,15 +215,19 @@ def _estimate_cassandra_cluster_zonal(
...
     # Netflix Cassandra doesn't like to deploy on really small instances
     if instance.cpu < 2 or instance.ram_gib < 14:
         return None
 
+    # temporarily dont suggest EBS instances
+    if instance.drive is None:
+        return None
+
     # if we're not allowed to use gp2, skip EBS only types
     if instance.drive is None and require_local_disks:
```
jolynch, line 232: can you just flip the default of require_local_disks to true?

tests/netflix/test_cassandra.py
```diff
@@ -78,70 +81,70 @@ def test_capacity_small_fast():
         assert small_result.cluster_params["cassandra.heap.table.percent"] == 0.11
 
 
-def test_ebs_high_reads():
```
jolynch, line 81: if you did it with the param then you can just pass required_local_disks false to these tests and keep the context where EBS _would_ be useful.

service_capacity_modeling/models/org/netflix/cassandra.py
ayushisingh29: Good point. I have fixed it along with the tests.

tests/netflix/test_cassandra.py
ramsrivatsa: That's a good point. However, I do want to preserve the old tests as this is a temporary change. And hence commenting out the tests could be a better choice?

ayushisingh29: If I set `"require_local_disks": False` and `"require_attached_disks": True, ` in the test then it will test for ebs case. So we are achieving the result of old tests just by passing extra params.

PR #97, opened 2024-11-27
Adding support for multiregion aurora

coleapz:


tests/netflix/test_postgres.py
```diff
@@ -99,7 +99,7 @@ def test_small_footprint_multi_region():
         desires=small_footprint,
         num_regions=3,
     )
-    assert cap_plan[0].candidate_clusters.zonal[0].instance.name == "m5d.xlarge"
+    assert cap_plan[0].candidate_clusters.regional[0].instance.name == "db.r5.large"
```
ShengweiWang, line 102: consider adding more validation on if the capacity planner is able to provide a multi-region plan

coleapz: after conversation we decided its okay

PR #98, opened 2025-01-08
Expose instances info through planner

shy-1234:


jolynch (commented):
Does the existing hardware_shapes property work?

PR #102, opened 2025-01-31
Add tool to fetch latest EC2 pricing from AWS API

jasonk000:
Introduces a tool to automatically fetch EC2 pricing and inject into the hardware pricing information for AWS. Also, update the hardware profile loading to allow multiple files, read by lexicographical sort.


service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_zz-overrides.json
```diff
@@ -0,0 +1,68 @@
+{
+  "us-east-1": {
+    "instances": {
+      "f5.large.deprecated": {"annual_cost": 300.0, "lifecycle": "deprecated"},
+      "m6id.xlarge":   {"annual_cost": 781.66},
+      "m6id.4xlarge":  {"annual_cost": 3126.64},
```
jasonk000, line 6: Just to call out, these appear to me to differ from the API fetch. Changing these to the current dynamically loaded values causes a failure on `tests/netflix/test_cassandra.py::test_capacity_high_writes`. I believe this is because it shifts the relative pricing, which shifts number of expected instances in the assertion.

@jolynch recommend if you can validate that, and we can fast-follow to remove these overrides.

jolynch:
> Also, update the hardware profile loading to allow multiple files, read by lexicographical sort.

service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_ec2.json
```diff
@@ -0,0 +1,2514 @@
...
+      "r5d.24xlarge": {
+        "annual_cost": 22766.33
+      },
+      "r6id.8xlarge": {
+        "annual_cost": 7971.0
+      },
+      "g5.4xlarge": {
+        "annual_cost": 5349.0
+      },
+      "c4.xlarge": {
```
jolynch, line 220: Can this file be sorted lexiographically? I think there is an option in json dump to write in source order and then we can just insert in lexiographic order.

service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_zz-overrides.json
jolynch: Yup I can own that, I'm 99% sure it's fine.

tools/fetch_pricing.py
```diff
@@ -0,0 +1,124 @@
...
+        project_root,
+        "service_capacity_modeling",
+        "hardware",
+        "profiles",
+        "pricing",
+        "aws",
+        "3yr-reserved_ec2.json",
+    )
+    with open(output_file, "w", encoding="utf-8") as f:
+        json.dump(output, f, indent=2)
```
jolynch, line 102: I think this is where we could sort

jolynch (approved):
lg2m, minor suggestions and I can fast follow if needed.

service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_ec2.json
jasonk000: Can do a sort... more of a meta discussion though:

My thinking was that different files can do different chunks of data, so we can add a _rds.json to store the rds pricing data. It's just the load stage that has to be able to blend the files programmatically.

If you wanted to insert shape data, with manual input, in different files, we could apply a similar treatment to the loading of shapes/aws.json data, and combine instance data. For ex, shapes/aws_sizes, shapes/aws_fio, shapes/aws_iperf, shapes/.... could all be manually maintained distinctly, and blended at load time.

An ideal end-state would be absolutely as much as possible should be generated by tooling, and a minimum amount of cognitive effort to apply manual inputs. The code should take care of loading and combining files correctly, rather than pushing the thinking to people manually typing it in.

All that is to say, I don't think this file should ever be modified by a human, so I'm not sure the value of sorting it. Maybe I've misunderstood! Let me know what you're thinking.

tools/fetch_pricing.py
jasonk000: fixed in b6fbf83

service_capacity_modeling/hardware/__init__.py
```diff
@@ -69,20 +83,58 @@ def price_hardware(hardware: Hardware, pricing: Pricing) -> GlobalHardware:
...
+
+    if isinstance(price_paths, str):
+        price_paths = [Path(price_paths)]
+
+    if isinstance(shape_path, str):
+        shape_path = Path(shape_path)
+
+    combined_pricing: Dict = {}
+
+    for price_path in price_paths:
```
shy-1234, line 134: Just want to double check: 
do we expect to have multiple files (more than 2) in pricing/aws in the future?
do we have expectation of which file being read first? (the latter would override the value from the previous, so the order seems matter. If with duplicate record, which one should be the truth? The one from API or the one hard-coded?)

service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_zz-overrides.json
jasonk000: I tidied up and added explicit tests on the sort. I had to comment out this test, to be fixed. It turns out the failing test is because of the sort order in the input!, I wasn't able to do both sort and have this test pass. In other words, I no longer think this is because of the relative pricing shift, it seems to be more the traversal order of instance type names.

service_capacity_modeling/hardware/__init__.py
jasonk000: > do we expect to have multiple files (more than 2) in pricing/aws in the future?

Yes, I see a `_ec2.json`, we can also likely add a `_rds.json` (for the db.xxx instance types), and `_ebs.json` if we build scripting for those. Whilst they could be in the same file, we have flexibility on how we want to combine them.

> do we have expectation of which file being read first? (the latter would override the value from the previous, so the order seems matter. If with duplicate record, which one should be the truth? The one from API or the one hard-coded?)

Yes, they are specifically loaded in-lexicographic-order from the file system. I named the file `_zz-overrides.json` to hopefully help maintainer see the ordering. My assumption/expectation is that any auto-generated pricing files would not overlap or even if they do, that the pricing would not be different.

shy-1234 (approved):

jolynch:
Will follow up on the test change.

PR #103, opened 2025-01-31
Start adding the gen 6 with ssd and automate a bit for 7th

jolynch:
Start of a tool for automatically creating shapes with @shy-1234 and @jasonk000 . I generated these shapes by doing r6id.2xlarge and then "reshaping" via

```
.tox/dev/bin/python -m service_capacity_modeling.tools.projection r6id.2xlarge
```

@shy-1234 I think the "pull_family" api in the tool is ultimately what we want to implement - I will work on a fast follow that pulls shapes and pricing from multiple files so you can focus on that integration for m7a.


service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved.json
```diff
@@ -51,6 +53,31 @@
       "m6id.2xlarge":  {"annual_cost": 1563.32},
       "m6id.4xlarge":  {"annual_cost": 3126.64},
       "m6id.8xlarge":  {"annual_cost": 6253.28},
+
```
jasonk000, line 56: No longer needed after the pricing Pr is merged (#102), but I think we are on the same page here.

service_capacity_modeling/interface.py
```diff
@@ -333,12 +335,43 @@ def family(self):
...
 
+    def merge_with(self, overrides: "Instance") -> "Instance":
+        self_dict = self.model_dump()
+        other_dict = overrides.model_dump(exclude_unset=True)
+
+        for (k, v) in other_dict.items():
+            # TODO we need a deep merge on drive (recursive merge)
+            if k in ("platforms",):
+                # Unique merge platforms
+                merged_platforms = list(self_dict.get("platforms", []))
```
jasonk000, line 364: how about

```
merged_platforms = list(
  set(self_dist.get("platforms", [])) | set(other_dict.get("platforms", []))
)
```

service_capacity_modeling/hardware/profiles/shapes/aws.json
```diff
@@ -587,6 +603,180 @@
...
+        "block_size_kib": 4,
+        "read_io_latency_ms": {
+          "low": 0.1,
+          "mid": 0.125,
+          "high": 0.17,
+          "confidence": 0.9,
+          "minimum_value": 0.05,
+          "maximum_value": 2.05
+        },
+        "annual_cost": 0.0
```
shy-1234, line 681: likely a bug in auto_shape. We shouldn't have annual_cost in generated shapes

service_capacity_modeling/tools/auto_shape.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def name(input_size: Fraction) -> str:
+        if input_size.denominator == 1 and input_size.numerator != 1:
+            size = str(input_size.numerator) + "xlarge"
+        else:
+            size = {
+                Fraction(1, 4): "medium",
+                Fraction(1, 2): "large",
+                Fraction(1, 1): "xlarge",
+            }[input_size]
```
jasonk000, line 67: Should we also handle the `.metal-NN` sizes? I'd forgotten this! Low priority.

jasonk000 (approved):
LGTM for expedience.

service_capacity_modeling/hardware/profiles/shapes/aws.json
shy-1234: hmm.. should have been excluded in reshape_family where we del the field.. Maybe this is from the previous result? it's not showing in the xlarge and large.

shy-1234: It's actually from the property in Drive class.. Not sure whether there is any non-tedious way to remove it.. Maybe we can manually remove it after reshape

service_capacity_modeling/interface.py
jolynch: I'm going to pull in the nflx-configs deep merge functionality when I add layering, which can deep merge arbitrary structured objects with override support.

service_capacity_modeling/tools/auto_shape.py
jolynch: Yeah metals are ... interesting ... let's take those separately

PR #108, opened 2025-02-06
Flesh out CurrentClusters in CapacityDesires input.

sdab:
We now have cpu, memory, and network utilization in the CurrentCapacity section of the CapacityDesires input. Furthermore the generate_scenarios() call which filters out instances now takes that shape into account.

This allows for callers of the plan api to (optionally) input the current cluster capacity and use that to inform instance selection. Models can furthermore use this to determine cluster size based on current utilziation.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -576,6 +576,58 @@ class CapacityPlanner:
         num_results = num_results or self._default_num_results
         return reduce_by_family(plans)[:num_results]
 
+    # Calculates the minimum cpu, memory, and network requirements based on desires.
+    def _per_instance_requirements(self, desires) -> Tuple[int, float, float]:
+
+        # Applications often set fixed reservations of heap or OS memory
+        per_instance_mem = (
+            desires.data_shape.reserved_instance_app_mem_gib
+            + desires.data_shape.reserved_instance_system_mem_gib
+        )
+
+        # Applications often require a minimum amount of true parallelism
+        per_instance_cores = int(
+            math.ceil(
+                desires.query_pattern.estimated_read_parallelism.mid
+                + desires.query_pattern.estimated_write_parallelism.mid
+            )
+        )
+
+        per_instance_net = 0.0
+
+        current_capacity = (
+            None
+            if desires.current_clusters is None
+            else (
+                desires.current_clusters.zonal[0]
+                if len(desires.current_clusters.zonal)
+                else desires.current_clusters.regional[0]
+            )
+        )
+        # Return early if we dont have current_capacity set.
+        if current_capacity is None or current_capacity.cluster_instance is None:
+            return (per_instance_cores, per_instance_mem, per_instance_net)
+
+        # Calculate CPU requirements based on current capacity
+        current_cpu_utilization = current_capacity.cpu_utilization.high / 100.0
+        current_instance_cpus = current_capacity.cluster_instance.cpu
+        required_cores = math.ceil(current_cpu_utilization * current_instance_cpus)
+        per_instance_cores = max(per_instance_cores, required_cores)
+
+        # Calculate memory requirements based on current capacity
+        current_memory_utilization_gib = (
+            current_capacity.memory_utilization_mib.high / 1024
+        )
+        per_instance_mem = max(per_instance_mem, current_memory_utilization_gib)
+
+        # Calculate network requirements based on current capacity
+        current_network_utilization_mbps = (
+            current_capacity.network_utilization_mbps.high
+        )
+        per_instance_net = max(per_instance_net, current_network_utilization_mbps)
+
+        return (per_instance_cores, per_instance_mem, per_instance_net)
+
     def generate_scenarios(  # pylint: disable=too-many-positional-arguments
         self,
         model,
@@ -598,19 +650,13 @@ class CapacityPlanner:
             num_regions=num_regions,
         )
 
-        # Applications often set fixed reservations of heap or OS memory, we
-        # should not even bother with shapes that don't meet the minimums
-        per_instance_mem = (
-            desires.data_shape.reserved_instance_app_mem_gib
-            + desires.data_shape.reserved_instance_system_mem_gib
-        )
-        # Applications often require a minimum amount of true parallelism
-        per_instance_cores = int(
-            math.ceil(
-                desires.query_pattern.estimated_read_parallelism.mid
-                + desires.query_pattern.estimated_write_parallelism.mid
-            )
-        )
+        # We should not even bother with shapes that don't meet the minimums
+        (
+            per_instance_cores,
+            per_instance_mem,
+            per_instance_net,
+        ) = self._per_instance_requirements(desires)
+
         allowed_platforms: Set[Platform] = set(model.allowed_platforms())
         allowed_drives: Set[str] = set(drives or [])
         for drive_name in model.allowed_cloud_drives():
@@ -635,6 +681,7 @@ class CapacityPlanner:
                 if (
                     per_instance_mem > instance.ram_gib
                     or per_instance_cores > instance.cpu
+                    or per_instance_net > instance.net_mbps
                 ):
                     continue
 
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -642,12 +642,6 @@ class CapacityPlanner:
             num_regions=num_regions,
         )
 
-        # We should not even bother with shapes that don't meet the minimums
-        (
-            per_instance_cores,
-            per_instance_mem,
-        ) = self._per_instance_requirements(desires)
-
         allowed_platforms: Set[Platform] = set(model.allowed_platforms())
         allowed_drives: Set[str] = set(drives or [])
         for drive_name in model.allowed_cloud_drives():
@@ -660,6 +654,12 @@ class CapacityPlanner:
 
         # Set current instance object if exists
         _set_instance_objects(desires, hardware)
+
+        # We should not even bother with shapes that don't meet the minimums
+        (
+            per_instance_cores,
+            per_instance_mem,
+        ) = self._per_instance_requirements(desires)
 
         if model.run_hardware_simulation():
             for instance in hardware.instances.values():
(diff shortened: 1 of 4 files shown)


service_capacity_modeling/interface.py
```diff
@@ -683,7 +683,12 @@ class CurrentClusterCapacity(ExcludeUnsetModel):
     cluster_instance_name: str
     cluster_instance: Optional[Instance] = None
     cluster_instance_count: Interval
+    # The distribution cpu utilization in the cluster.
```
sdab, line 686: Im open to some api design advice here. I think per-node distributions are enough, but we could also do cluster wide utilization. Im thinking with just per-node we can estimate that by multiplying by number of instances.

service_capacity_modeling/interface.py
```diff
@@ -683,7 +683,12 @@ class CurrentClusterCapacity(ExcludeUnsetModel):
     cluster_instance_name: str
     cluster_instance: Optional[Instance] = None
```
sdab, line 684: I think we should either get rid of this field and look it up from our shapes + cluster_instance_name or dont have it as optional.

service_capacity_modeling/capacity_planner.py
```diff
@@ -576,6 +576,58 @@ def _plan_certain(  # pylint: disable=too-many-positional-arguments
...
+                if len(desires.current_clusters.zonal)
+                else desires.current_clusters.regional[0]
+            )
+        )
+        # Return early if we dont have current_capacity set.
+        if current_capacity is None or current_capacity.cluster_instance is None:
+            return (per_instance_cores, per_instance_mem, per_instance_net)
+
+        # Calculate CPU requirements based on current capacity
+        current_cpu_utilization = current_capacity.cpu_utilization.high / 100.0
```
ramsrivatsa, line 612: I think this might be almost and always not used. 

Linux tools report average CPU utilization, which is generally low for stateful datastores due to I/O activity. 

Hence, if we are using these, we need to override them in the tech-specific models. Should'nt hurt to have this.

ramsrivatsa: In cassandra we override it here. 
https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/b1d140cdf60c5ded332a98fd8899d7a3f068a2b9/service_capacity_modeling/models/org/netflix/cassandra.py#L97-L116

service_capacity_modeling/interface.py
ramsrivatsa: That might require a some rewrites of this code across all datastores. There are places where we translate from cluster -> regional and back again. 

Good team project idea for a single day hackathon.

service_capacity_modeling/capacity_planner.py
sdab: Right, I think the goal of generate_scenarios is to filter out obviously wrong instance types. Models can then reject more strictly based on more context.

service_capacity_modeling/capacity_planner.py
```diff
@@ -576,6 +576,58 @@ def _plan_certain(  # pylint: disable=too-many-positional-arguments
...
+        current_memory_utilization_gib = (
+            current_capacity.memory_utilization_mib.high / 1024
+        )
+        per_instance_mem = max(per_instance_mem, current_memory_utilization_gib)
+
+        # Calculate network requirements based on current capacity
+        current_network_utilization_mbps = (
+            current_capacity.network_utilization_mbps.high
+        )
+        per_instance_net = max(per_instance_net, current_network_utilization_mbps)
```
jolynch, line 627: Does per node network make sense as a filter? I get memory (true usage, not allocated) and cores, but network is not generally in-compressible. I think as long as the cluster network capacity is high enough we could assume load balancing will resolve this?

service_capacity_modeling/capacity_planner.py
```diff
@@ -576,6 +576,58 @@ def _plan_certain(  # pylint: disable=too-many-positional-arguments
...
+
+        # Calculate CPU requirements based on current capacity
+        current_cpu_utilization = current_capacity.cpu_utilization.high / 100.0
+        current_instance_cpus = current_capacity.cluster_instance.cpu
+        required_cores = math.ceil(current_cpu_utilization * current_instance_cpus)
+        per_instance_cores = max(per_instance_cores, required_cores)
+
+        # Calculate memory requirements based on current capacity
+        current_memory_utilization_gib = (
+            current_capacity.memory_utilization_mib.high / 1024
```
jolynch, line 619: Similar question here, we just have to be careful to measure actually resident memory and not allocated (e.g. Java prior to ZGC will just show 80% of the box "used").

service_capacity_modeling/interface.py
jolynch: Yeah I think we should just do name, agreed

service_capacity_modeling/capacity_planner.py
```diff
@@ -576,6 +576,58 @@ def _plan_certain(  # pylint: disable=too-many-positional-arguments
...
+            )
+        )
+        # Return early if we dont have current_capacity set.
+        if current_capacity is None or current_capacity.cluster_instance is None:
+            return (per_instance_cores, per_instance_mem, per_instance_net)
+
+        # Calculate CPU requirements based on current capacity
+        current_cpu_utilization = current_capacity.cpu_utilization.high / 100.0
+        current_instance_cpus = current_capacity.cluster_instance.cpu
+        required_cores = math.ceil(current_cpu_utilization * current_instance_cpus)
```
jolynch, line 614: This seems odd to me, since the model can return a higher number of smaller computer they could get to this utilization number a different way (e.g. 2x computers would halve the utilization)

jolynch (changes requested):
I feel like we're taking a cluster level input and then filtering out node level questions, I think we want to just pass these through to the model to short-circuit requirement generation (but more small shapes is still viable for cpu and network, min ram is kinda incompressible)

service_capacity_modeling/capacity_planner.py
ramsrivatsa: Might not be applicable for EVCache --> Memcached as there is no JVM. 

Again, we are under the assumption that apps will override them. However, @jolynch, is your concern that you want to cap memory at probably around 80% CPU?

sdab: Right but how the caller fills in memory_utilization is up to them. i.e. if a jvm app, they can pass in resident.

service_capacity_modeling/capacity_planner.py
sdab: On Titus we generally needed some per-container guaranteed network bw, but thats due to multi-tenancy. You are probably right that its enough for the cluster to have enough network.

sdab: Removed network as a filter.

service_capacity_modeling/interface.py
sdab: Actually I take this back. generate_scenarios fills this in rather than the caller (probably why its optional). I do see that this would be useful since the model doesnt have access to the hardware shapes. I did however move my helper call until after generate_scenarios fills in the field.

sdab:
> I feel like we're taking a cluster level input and then filtering out node level questions, I think we want to just pass these through to the model to short-circuit requirement generation (but more small shapes is still viable for cpu and network, min ram is kinda incompressible)

It is a per-cluster node distribution utilization. i.e. these are node distributions

service_capacity_modeling/capacity_planner.py
jolynch: Hm I stilll think this is a tad odd to filter on this, since we could achieve the utilization with fewer cores (but more machines) - but I do agree it's conservative as when rightsizing we would never go up in per machine utilized cores

So if you have 100 4xl with 50% utilization, thats 16 * 8 minimum 8 vcpus, imo it would be fine to have 200 2xl (which only have 8 cores total per box

jolynch (approved):
I'm still a bit skeptical on the filtering part, but let's get the API out and play with the implementation.

service_capacity_modeling/capacity_planner.py
sdab: @jolynch I think this calculation adheres to your example. The filter would indeed resolve to min 8vcpus which would indeed allow 2xls through. It would filter out instances with less than 8vcpus.

sdab: One potential issue is that we may need to normalize these, but Ill leave that for a follow up.

jasonk000: I think a more extreme example. If we have a current cluster of 100 * m5.8xl @ 80%, and scale linearly, it could theoretically run 800 * m5.xl @ 80%. If I read correctly, I think this would come up with `current_util 80% * current_cpus = 32 = 25cores`, which would exclude anything below 8xl IIUC?

sdab: Hmm yes that does help. I think my mental model is incorrect, I have indeed been treating cpus as "incompressible" as Joey likes to put it.

It does seem wrong to think we could go to the extreme and run on a bunch of 1vcpu instances, but my calculation here is likely much too strict. Maybe I should remove it here and just leave it up to the model then.

service_capacity_modeling/interface.py
```diff
@@ -683,7 +683,12 @@ class CurrentClusterCapacity(ExcludeUnsetModel):
...
     cluster_instance_count: Interval
+    # The distribution cpu utilization in the cluster.
     cpu_utilization: Interval
+    # The per node distribution of memory used in mib.
+    memory_utilization_mib: Interval
```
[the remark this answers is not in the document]
sdab: Done

service_capacity_modeling/capacity_planner.py
sdab: Removed filtering on cpu utilization.

service_capacity_modeling/capacity_planner.py
```diff
@@ -624,6 +647,12 @@ def generate_scenarios(  # pylint: disable=too-many-positional-arguments
         # Set current instance object if exists
         _set_instance_objects(desires, hardware)
 
+        # We should not even bother with shapes that don't meet the minimums
+        (
+            per_instance_cores,
+            per_instance_mem,
+        ) = self._per_instance_requirements(desires)
```
argha-c, line 654: nit: Does it make sense to materialize this tuple in a model? The composite feels domain relevant.

sdab: hmm perhaps? This was there before, it was just specific to stateful desires (i.e. read_threads -> cpus). I generalized it here. Since it was in use before, I imagine that in practice most models agreed on this conversion.

Additionally, models have a way to bypass this completely.

PR #109, opened 2025-02-06
Capacity planner for GraphKV abstraction

kartik894:
This PR adds a simple capacity model for GraphKV, KV and Cassandra. In this V1 version, we add a custom hardcoded amplification factor read and write requests per second for KV. In the following versions, we will focus on building custom graphkv configs that we can leverage for deriving a better estimate of the amplification factor.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/graphkv.py
+++ b/service_capacity_modeling/models/org/netflix/graphkv.py
@@ -1,0 +1,206 @@
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Optional
+from typing import Tuple
+
+from .stateless_java import nflx_java_app_capacity_model
+from service_capacity_modeling.interface import AccessConsistency
+from service_capacity_modeling.interface import AccessPattern
+from service_capacity_modeling.interface import CapacityDesires
+from service_capacity_modeling.interface import CapacityPlan
+from service_capacity_modeling.interface import Consistency
+from service_capacity_modeling.interface import DataShape
+from service_capacity_modeling.interface import Drive
+from service_capacity_modeling.interface import FixedInterval
+from service_capacity_modeling.interface import GlobalConsistency
+from service_capacity_modeling.interface import Instance
+from service_capacity_modeling.interface import Interval
+from service_capacity_modeling.interface import QueryPattern
+from service_capacity_modeling.interface import RegionContext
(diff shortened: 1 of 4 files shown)


service_capacity_modeling/models/org/netflix/graphkv.py
```diff
@@ -0,0 +1,206 @@
...
+                    estimated_state_size_gib=Interval(
+                        low=10, mid=50, high=200, confidence=0.98
+                    ),
+                    reserved_instance_app_mem_gib=8,
+                ),
+            )
+        else:
+            return CapacityDesires(
+                query_pattern=QueryPattern(
+                    access_pattern=AccessPattern.latency,
```
abersnaze, line 156: is this the right access_pattern? its hard to tell what is the difference between these two branches

tox.ini
```diff
@@ -39,6 +39,7 @@ deps =
     isodate
     setuptools
     boto3
+    pytest
```
abersnaze, line 42: i don't think this is needed. it isn't running unit tests as part of pre-commit.

kartik894: Discussed offline: It's required by test_graphkv.py so that we can run parameterized tests. The linter checks the tests/ directory, hence I added it as a dependency for the pre-commit hook.

service_capacity_modeling/models/org/netflix/graphkv.py
```diff
@@ -0,0 +1,206 @@
...
+    def compose_with(
+        user_desires: CapacityDesires, extra_model_arguments: Dict[str, Any]
+    ) -> Tuple[Tuple[str, Callable[[CapacityDesires], CapacityDesires]], ...]:
+        def _modify_kv_desires(
+            user_desires: CapacityDesires,
+        ) -> CapacityDesires:
+            relaxed = user_desires.model_copy(deep=True)
+
+            # TODO: introduce a custom config file for graphkv
+            default_amplification = 10
```
kelebra, line 66: What do you think about starting with 2 parameters: max depth (2) and properties (5). Properties will impact KV storage and RPS / WPS; depth will impact RPS / WPS only.

kelebra (approved):

service_capacity_modeling/models/org/netflix/graphkv.py
kartik894: Done.

PR #110, opened 2025-02-10
Adds headroom estimator notebook and simple service headroom calculator

jasonk000:


service_capacity_modeling/simple/instances.py
```diff
@@ -0,0 +1,61 @@
+from typing import Optional
+from service_capacity_modeling.hardware import shapes
+
+__all__ = ["get_simple_instance_headroom_target"]
```
sdab, line 4: Does this do anything in this file?

jasonk000: I honestly do not know. Cargo culted.

service_capacity_modeling/simple/instances.py
```diff
@@ -0,0 +1,61 @@
...
+    # We estimate each virtual core to be about 60% as effective as a physical core
+    HT_PENALTY = 0.6
+
+    # Adjust effective cores if hyperthreading is enabled
+    # This accounts for the reduced effectiveness of virtual cores
+    if is_ht:
+        cores = cores * HT_PENALTY
+
+    # Calculate required headroom using magic formula:
+    return 0.712 / (cores**0.448)
```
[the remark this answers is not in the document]
jasonk000: Yes, I referred to the notebook:

[line 14](https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/110/files/f9099ffbecbe53ed75e5c92811e9020ade879018#diff-bb797bd4d0a2daa27ce43f4c449ade4135fe5e5d6bc4c82578cba3e03bd5bebbR14) and [line 53](https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/110/files/f9099ffbecbe53ed75e5c92811e9020ade879018#diff-bb797bd4d0a2daa27ce43f4c449ade4135fe5e5d6bc4c82578cba3e03bd5bebbR53)

Did you mean you wanted a more concrete description here?

service_capacity_modeling/simple/instances.py
```diff
@@ -0,0 +1,61 @@
...
+    The headroom target should be the percentage of CPU that should be
+    reserved for headroom to ensure sensible performance profile.
+
+    This could be 1-utilization_target, however we leave the ultimate
+    utilization_target to the caller, since, we do not know how much
+    operational headroom they want to leave (ie: success buffer).
+
+    For example, a response here of "headroom = 15%", means caller could
+    decide with a success_buffer=1 to use a utilization_target of 85%.
+    For success_buffer>1, they should target below 85% utilization.
```
sdab, line 48: Is it worth a second helper function that accepts a success_buffer parameter? I know its just division, but still.

service_capacity_modeling/simple/instances.py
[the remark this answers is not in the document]
sdab: Ah I had to go into preview mode in the branch, not the PR, to see the notebook. It does indeed explain where these come from.

notebooks/headroom-estimator.ipynb
sdab, line 1: Jason, if I am interpreting correctly, this model implies that we can run larger nodes hotter while still adhering to success buffers correct? Does this further imply that bigger nodes are better/more cost efficient (needing less overall headroom and thus less overall cpus) for apps that can scale vertically?

jasonk000: Yes, that's precisely correct.

service_capacity_modeling/simple/instances.py
jasonk000: removed in 3b51a7f

service_capacity_modeling/simple/instances.py
jasonk000: Based on discussion w/ Felipe and Joey they want to make that decision on their end. I think it's better to leave the concepts un-mixed as we then start bringing in an assumption of "how" we calculate any operational headroom.

sdab (approved):

service_capacity_modeling/simple/instances.py
```diff
@@ -0,0 +1,60 @@
...
+    # Calculate required headroom using magic formula:
+    return 0.712 / (cores**0.448)
+
+
+def _is_instance_type_hyperthreads(instance_type: str) -> bool:
+    instance_family = _family_from_instance_type(instance_type)
+    return instance_family not in FULL_CORE_INSTANCE_FAMILIES
+
+
+def get_simple_instance_headroom_target(instance_type: str) -> Optional[float]:
```
felipems, line 34: tangential, but in the future, would this be an enum or a class?

service_capacity_modeling/simple/instances.py
```diff
@@ -0,0 +1,60 @@
...
+from service_capacity_modeling.hardware import shapes
+
+FULL_CORE_INSTANCE_FAMILIES = ["c7a", "m7a", "r7a"]
+
+
+def _family_from_instance_type(instance_type: str) -> str:
+    return instance_type.split(".")[0]
+
+
+def _headroom_approx(cores, is_ht):
```
felipems, line 12: nit: types

service_capacity_modeling/simple/instances.py
felipems: I feel like if this is gonna be used as a lib it would be nice to be able to import instance types and have that client be able to do instance type things.

felipems (commented):
Nothing major to complain about :D 
Mostly my thoughts are on how we plan to use an instance library and whether that can be an object I can use.

felipems (approved):

service_capacity_modeling/simple/instances.py
sdab: This library has an instance type, you can see it looked up by in this function. It does generally key by string/name though

service_capacity_modeling/simple/instances.py
jasonk000: thanks, fixed in 3676fa8

service_capacity_modeling/simple/instances.py
jasonk000: Good idea. Implemented in 3676fa8 -- I created

- created `get_simple_instance_headroom` which takes an `Instance` which is used in the rest of the lib

- and the helper `get_simple_instance_headroom_target_by_name` to use the `instance_type: str` by name

PR #111, opened 2025-02-10
allow other types of intervals

abersnaze:


service_capacity_modeling/interface.py
```diff
@@ -623,10 +619,10 @@ class QueryPattern(ExcludeUnsetModel):
     # to provision such that we don't involve oncall
     # Note that these summary statistics will be used to create reasonable
     # distribution approximations of these operations (yielding p25, p99, etc)
-    read_latency_slo_ms: FixedInterval = FixedInterval(
+    read_latency_slo_ms: Interval = FixedInterval(
```
jolynch, line 622: @abersnaze this probably doesn't have the effect you want, this will make it so we _simulate_ this interval during plan uncertain.

PR #112, opened 2025-02-11
Add disk utilization to CurrentClusterCapacity input

sdab:
Adds disk_utilization_gib to the CurrentClusterCapacity which I think Rams work will need.

I also added some default utilization values (set to 0) so callers dont have to set them.

### DIFF
--- a/service_capacity_modeling/interface.py
+++ b/service_capacity_modeling/interface.py
@@ -680,11 +680,13 @@ class CurrentClusterCapacity(ExcludeUnsetModel):
     cluster_instance: Optional[Instance] = None
     cluster_instance_count: Interval
     # The distribution cpu utilization in the cluster.
-    cpu_utilization: Interval
+    cpu_utilization: Interval = certain_float(0.0)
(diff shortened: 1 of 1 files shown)


service_capacity_modeling/interface.py
```diff
@@ -680,11 +680,13 @@ class CurrentClusterCapacity(ExcludeUnsetModel):
     cluster_instance: Optional[Instance] = None
     cluster_instance_count: Interval
     # The distribution cpu utilization in the cluster.
-    cpu_utilization: Interval
+    cpu_utilization: Interval = certain_float(0.0)
```
jasonk000, line 683: Would this be better as Optional[Interval]?

sdab: I was debating it. Optional is a little annoying with having to be checked when reading it. I was thinking the defaults could be used safely.

ramsrivatsa (approved):

ayushisingh29 (approved):

PR #116, opened 2025-02-17
Make buffers more flexible to support all needs

jolynch:
I've made a few changes to the Buffers interface that hopefully will allow rightsizing and scaling for load spikes use cases without requiring mapping through buffer setups. This was based on feedback from @ayushisingh29 , @ramsrivatsa, @sdab , @argha-c , @jasonk000  and @felipems. Main changes:

* Instead of a simple float buffers are now a `Buffer` object which contains a ratio and components.
* The `buffer_for_components` logic is now just filtering any buffers which match that component
* Added `Buffer.intent` to allow for rightsizing to collapse buffers to desired, or preserve them as needed.

### DIFF
--- a/service_capacity_modeling/interface.py
+++ b/service_capacity_modeling/interface.py
@@ -770,112 +770,81 @@ class BufferComponent(str, Enum):
     the Buffers interface itself (should be str).
     """
 
-    # "Traffic" related buffers, e.g. CPU and Network
+    # [Query Pattern] a.k.a. "Traffic" related buffers, e.g. CPU and Network
     compute = "compute"
-    # "Dataset" related buffers, e.g. Disk and Memory
+    # [Data Shape]    a.k.a. "Dataset" related buffers, e.g. Disk and Memory
     storage = "storage"
-    # Specific resources
+
+    # Resource specific component
     cpu = "cpu"
     network = "network"
     disk = "disk"
-    # Memory can be used for various buffers
     memory = "memory"
-    memory_traffic = "memory-traffic"
-    memory_write_cache = "memory-write-cache"
-    memory_read_cache = "memory-read-cache"
+
+
+class BufferIntent(str, Enum):
+    # Most buffers show "desired" buffer, this is the default
+    desired = "desired"
+    # ratio on top of existing buffers to ensure exists. Generally combined
+    # with a different desired buffer to ensure we don't just scale needlessly
+    scale = "scale"
+    # Ignore model preferences, just preserve existing buffers
+    preserve = "preserve"
 
 
 class Buffer(ExcludeUnsetModel):
-    # The type of buffer, intentionally not BufferComponent to allow
-    # models to define their own.
-    component: str
     # The value of the buffer expressed as a ratio over "normal" load e.g. 1.5x
-    ratio: float
-    # Set only if the source of the buffer is not the direct component
-    source: Optional[str] = None
+    ratio: float = 1.0
+    # What is the intent of this buffer directive, almost always is desired
+    intent: BufferIntent = BufferIntent.desired
+    # The components of buffer this influences, almost always is "compute" (IPC success)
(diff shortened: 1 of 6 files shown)


service_capacity_modeling/interface.py
```diff
@@ -770,112 +770,76 @@ class BufferComponent(str, Enum):
...
     network = "network"
     disk = "disk"
-    # Memory can be used for various buffers
     memory = "memory"
-    memory_traffic = "memory-traffic"
-    memory_write_cache = "memory-write-cache"
-    memory_read_cache = "memory-read-cache"
+
+
+class BufferIntent(str, Enum):
```
sdab, line 785: You might want some comments describing what these mean.

jolynch: Yeah I think I'll document more when I implement the different intents, just wanted to get eyes on the API a bit (and I do document a bit in the derived section below).

service_capacity_modeling/interface.py
```diff
@@ -770,112 +770,76 @@ class BufferComponent(str, Enum):
...
-    There is a series of fallbacks here over common desired buffers:
-        compute -> default
-        cpu -> compute -> default
-        network -> compute -> default
-        disk -> storage -> default
-        memory -> storage -> default
-        memory-traffic -> memory -> default
-        memory-write-cache -> memory -> default
-        memory-read-cache -> memory -> default
+    In this case when we query the buffer_for_components(components=["cpu"])
```
sdab, line 825: This makes sense for combining buffers.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -114,19 +124,18 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
         reference_shape = current_capacity.cluster_instance
     else:
         # We have no existing utilization to go from
-        # Keep half of the cores free for background work (compaction, backup, repair).
-        cpu_buffer = desires.buffers.buffer_for_component("cpu").ratio
         reference_shape = desires.reference_shape
-        needed_cores = math.ceil(sqrt_staffed_cores(desires) * cpu_buffer)
+        # Keep half of the cores free for background work (compaction, backup, repair).
+        needed_cores = math.ceil(sqrt_staffed_cores(desires) * background_buffer.ratio)
```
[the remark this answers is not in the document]
jolynch: I think that's the ultimate goal!  Currently most of the stateful models use sqrt_staffed_cores with an aggressive QoS parameter to produce both the success buffer.compute and failure buffer (cpu headroom), I need to refactor those to instead staff based on compute buffer (success buffer) and instance headroom (failure buffer) plus any additional CPU buffers - this is just a half way point that starts making those internal model buffers _visible_ to the buffer_for_components math (that's why source is returned).

service_capacity_modeling/interface.py
jolynch: Added some comments.

service_capacity_modeling/models/org/netflix/cassandra.py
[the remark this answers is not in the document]
sdab: ah ok thanks that helps me understand. Ok so to go further, this is basically ignoring the compute buffer and just computing its own thing based on the background_buffer ratio right?

sdab:
My only real API question is around the role of buffer "components". Here is what I think I understand:

- there are components for each low level resource (cpu, memory, disk, network, etc)
- there are also higher level components like "compute" which represent a combination of lower level components (i.e. cpu + mem)
- I expect a user/caller to set buffers in terms of high level components ("compute")
- I expect models to use (mostly) buffers in terms of low level components ("cpu", "mem", etc) which are calculated from the high level ones via buffer_for_components()

Do I have that right? Some explanation around that might be useful or maybe it will just be clear when we have more examples of usage.

PR #117, opened 2025-02-18
Minor refactoring for cores boost + effective cpu headroom

argha-c:
- refactors to indicate cpu boost w/ and w/o hyperthreading
- makes headroom calculation flexible. Defaults to queuing theory based headroom calculation.

### DIFF
--- a/service_capacity_modeling/models/common.py
+++ b/service_capacity_modeling/models/common.py
@@ -118,14 +118,14 @@ def normalize_cores(
     return max(1, math.ceil(core_count / (target_speed / reference_speed)))
 
 
-def _headroom_approx(cores: int, cores_boost: float = 1.0) -> float:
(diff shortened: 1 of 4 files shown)


service_capacity_modeling/models/common.py
```diff
@@ -118,14 +118,14 @@ def normalize_cores(
     return max(1, math.ceil(core_count / (target_speed / reference_speed)))
 
 
-def _headroom_approx(threads: int, thread_adj: float = 1.0) -> float:
+def _headroom_approx(cores: int, cores_boost: float = 1.0) -> float:
```
jolynch, line 121: Could we compromise on `cpu` to match the hardware language? These aren't cores (they are threads)

jolynch (approved):

PR #118, opened 2025-02-21
Revert "allow other types of intervals"

mckeealex:
Currently, if a user does not pass in `read_latency_slo_ms` or `write_latency_slo_ms` to `desires`, the following chain of events will occur in plan uncertain

1. `read/write_latency_slo_ms` will be set to `FixedInterval(low=0.4, mid=4, high=10, confidence=0.98)` from the `QueryPattern` defaults
2. The `read/write_latency_slo_ms` fields will then be filtered out when `desires_dict.model_dump()` is called during [`merge_with`](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/7f4d537a9167940ad7ee4b16ffacf176d7477d45/service_capacity_modeling/capacity_planner.py#L843) and then replaced with the default desires from [models/org/netflix/rds.py](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/b5ef07b5e6206a517373ac6a62edc3982775bea6/service_capacity_modeling/models/org/netflix/rds.py#L278) below
    - `read_latency_slo_ms=FixedInterval(
                    minimum_value=10,
                    maximum_value=2000,
                    low=300,
                    mid=800,
                    high=2000,
                    confidence=0.90
                ),`
3. The combination of defaults above then leads to `high` being set to 1 in `_beta_dist_from_interval` and therefore an [assertion error](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/5d0368a607577430b9cef12ad4326f62916ad03d/service_capacity_modeling/stats.py#L93) in `_beta_cost_fn_from_params`

The only changes I made here was swapping the type of `read_latency_slo_ms`/`write_latency_slo_ms` back to `FixedInterval` which changes it back to the previous functionality of the QueryPattern defaults being used and therefore no error

However, I feel like the previous behavior of `QueryPattern` defaults being used instead of the `netflix/rds.py` was probably not correct and that @abersnaze change may have uncovered this silent bug. If it is the case that `netflix/rds.py` is what should be defaulted to when no `read`/`write_latency_slo_ms` is passed into desires, then this code should be left as is and `netflix/rds.py` should be updated with defaults that don't trigger the `AssertionError`

But to be clear a have incredibly little context on this code and all of these points could be wildly off 😅

### DIFF
--- a/service_capacity_modeling/interface.py
+++ b/service_capacity_modeling/interface.py
@@ -248,8 +248,10 @@ class Drive(ExcludeUnsetModel):
 
     # These defaults are assuming a cloud SSD like a gp2 volume
     # If you disagree please change them in your hardware description
-    read_io_latency_ms: Interval = FixedInterval(low=0.8, mid=1, high=2, confidence=0.9)
-    write_io_latency_ms: Interval = FixedInterval(
+    read_io_latency_ms: FixedInterval = FixedInterval(
+        low=0.8, mid=1, high=2, confidence=0.9
+    )
+    write_io_latency_ms: FixedInterval = FixedInterval(
         low=0.6, mid=2, high=3, confidence=0.9
     )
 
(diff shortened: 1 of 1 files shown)


jolynch:
Yeah I think we should revert the entirety of @abersnaze 's interval change, I flagged that before I was concerned about it.

mckeealex:
@jolynch Yeah I agree, it also seems to be [breaking](https://radar.prod.netflix.net/apps/antigravity.capacity/prod/issues?states=unresolved%2Cregression&app=antigravity.capacity&id=b5245dbc5e4d21a3146fcefdb137c38a_v3&eventId=a3fbab7dcf54403dbeecd515bc74f503&ts=1740175077378) capacity planning for elasticsearch

akhaku:
Thanks, Alex! Might as well turn this PR into a straight-up `git revert 965b92df1b6ba1b1562f2165ddaf23aff00c52d5`

jolynch:
Thank you @mckeealex , @abersnaze let's pair on Monday to fix your round-trip issue a different way.

PR #126, opened 2025-03-15
use current cluster CPU utilization to compute needed_cores in kafka …

nickmahilani:
…model


service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -43,12 +43,25 @@ class ClusterType(str, Enum):
     ha = "high-availability"
 
 
+def target_cpu_utilzation(tier: int) -> float:
```
ramsrivatsa, line 46: r:utilzation:utilization

PR #127, opened 2025-03-17
common zonal requirement for a cluster using current capacity and buffers

ayushisingh29:
If the intent is scale, calculate the new required capacity. If new capacity > current then scale up the cluster else do nothing and return the current capacity. Cluster will not be right sized (scale down) in this case. 
If the intent is preserve, return the current capacity.

### DIFF
--- a/service_capacity_modeling/models/common.py
+++ b/service_capacity_modeling/models/common.py
@@ -694,50 +694,66 @@ def merge_plan(
     )
 
 
-def derived_buffer_for_component(buffer, components):
-    scale = 0
+def derived_buffer_for_component(buffer: Dict[str, Buffer], components: List[str]):
+    scale = 0.0
     preserve = False
 
-    if buffer is not None:
-        for bfr in buffer.values():
-            if any(i in components for i in bfr.components):
-                if bfr.intent == BufferIntent.scale:
-                    scale = max(scale, bfr.ratio)
-                if bfr.intent == BufferIntent.preserve:
-                    preserve = True
+    if not buffer:
+        return scale, preserve
+
+    for bfr in buffer.values():
+        if any(component in components for component in bfr.components):
+            if bfr.intent == BufferIntent.scale:
+                scale = max(scale, bfr.ratio)
+            if bfr.intent == BufferIntent.preserve:
+                preserve = True
 
     return scale, preserve
 
 
-def get_cores_with_buffer(current_capacity, buffers, instance):
+def get_cores_from_current_capacity(
+    current_capacity: CurrentClusterCapacity, buffers: Buffers, instance: Instance
+):
     # compute cores required per zone
     cpu_success_buffer = (1 - cpu_headroom_target(instance, buffers)) * 100
     current_cpu_utilization = current_capacity.cpu_utilization.mid
-    current_cores = (
-        current_capacity.cluster_instance.cpu
-        * current_capacity.cluster_instance_count.mid
-    )
+
+    if current_capacity.cluster_instance is None:
+        cluster_instance = shapes.instance(current_capacity.cluster_instance_name)
+    else:
+        cluster_instance = current_capacity.cluster_instance
+
+    current_cores = cluster_instance.cpu * current_capacity.cluster_instance_count.mid
 
     scale, preserve = derived_buffer_for_component(buffers.derived, ["compute", "cpu"])
-
+    # Scale and preserve for the same component should not be passed together.
+    # If user passes it, then scale will be preferred over preserve.
     if scale > 0:
         # if the new cpu core is less than the current,
         # then take no action and return the current cpu cores
         new_cpu_utilization = current_cpu_utilization * scale
         core_scale_up_factor = max(1.0, new_cpu_utilization / cpu_success_buffer)
         return math.ceil(current_cores * core_scale_up_factor)
-    elif preserve:
+
+    if preserve:
         return current_cores
-    else:
-        return int(current_cores * (current_cpu_utilization / cpu_success_buffer))
+
+    return int(current_cores * (current_cpu_utilization / cpu_success_buffer))
 
 
-def get_memory_with_buffer_gib(current_capacity, buffers):
+def get_memory_from_current_capacity(
+    current_capacity: CurrentClusterCapacity, buffers: Buffers
+):
     # compute memory required per zone
     current_memory_utilization = current_capacity.memory_utilization_gib.mid
-    current_ram_allocated = (
-        current_capacity.cluster_instance.ram_gib
-        * current_capacity.cluster_instance_count.mid
+
+    if current_capacity.cluster_instance is None:
+        cluster_instance = shapes.instance(current_capacity.cluster_instance_name)
+    else:
+        cluster_instance = current_capacity.cluster_instance
+
+    zonal_ram_allocated = (
+        cluster_instance.ram_gib * current_capacity.cluster_instance_count.mid
     )
 
     # These are the desired buffers
@@ -748,26 +764,35 @@ def get_memory_with_buffer_gib(current_capacity, buffers):
     scale, preserve = derived_buffer_for_component(
         buffers.derived, ["memory", "storage"]
     )
-
+    # Scale and preserve for the same component should not be passed together.
+    # If user passes it, then scale will be preferred over preserve.
     if scale > 0:
         # if the new required memory is less than the current,
         # then take no action and return the current ram
         return max(
             current_memory_utilization * scale * memory_buffer.ratio,
-            current_ram_allocated,
+            zonal_ram_allocated,
         )
-    elif preserve:
-        return current_ram_allocated
-    else:
-        return current_memory_utilization * memory_buffer.ratio
+
+    if preserve:
+        return zonal_ram_allocated
+
+    return current_memory_utilization * memory_buffer.ratio
 
 
-def get_network_with_buffer_mbps(current_capacity, buffers):
+def get_network_from_current_capacity(
+    current_capacity: CurrentClusterCapacity, buffers: Buffers
+):
     # compute network required per zone
     current_network_utilization = current_capacity.network_utilization_mbps.mid
-    current_network_allocated = (
-        current_capacity.cluster_instance.net_mbps
-        * current_capacity.cluster_instance_count.mid
+
+    if current_capacity.cluster_instance is None:
+        cluster_instance = shapes.instance(current_capacity.cluster_instance_name)
+    else:
+        cluster_instance = current_capacity.cluster_instance
+
+    zonal_network_allocated = (
+        cluster_instance.net_mbps * current_capacity.cluster_instance_count.mid
     )
 
     # These are the desired buffers
@@ -778,25 +803,37 @@ def get_network_with_buffer_mbps(current_capacity, buffers):
     scale, preserve = derived_buffer_for_component(
         buffers.derived, ["compute", "network"]
     )
-
+    # Scale and preserve for the same component should not be passed together.
+    # If user passes it, then scale will be preferred over preserve.
     if scale > 0:
         # if the new required network is less than the current,
         # then take no action and return the current bandwidth
         return max(
             current_network_utilization * scale * network_buffer.ratio,
-            current_network_allocated,
+            zonal_network_allocated,
         )
-    elif preserve:
-        return current_network_allocated
-    else:
-        return current_network_utilization * network_buffer.ratio
+
+    if preserve:
+        return zonal_network_allocated
+
+    return current_network_utilization * network_buffer.ratio
 
 
-def get_disk_with_buffer_gib(current_capacity, buffers):
+def get_disk_from_current_capacity(
+    current_capacity: CurrentClusterCapacity, buffers: Buffers
+):
     # compute disk required per zone
     current_disk_utilization = current_capacity.disk_utilization_gib.mid
-    current_disk_allocated = (
-        current_capacity.cluster_instance.drive.max_size_gib
+
+    if current_capacity.cluster_instance is None:
+        cluster_instance = shapes.instance(current_capacity.cluster_instance_name)
+    else:
+        cluster_instance = current_capacity.cluster_instance
+
+    assert cluster_instance.drive is not None, "Drive should not be None"
+
+    zonal_disk_allocated = (
+        cluster_instance.drive.max_size_gib
         * current_capacity.cluster_instance_count.mid
     )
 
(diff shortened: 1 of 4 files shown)


service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,153 @@ def merge_plan(
...
+                    scale = max(scale, bfr.ratio)
+                if bfr.intent == BufferIntent.preserve:
+                    preserve = True
+
+    return scale, preserve
+
+
+def get_cores_with_buffer(current_capacity, buffers, instance):
+    # compute cores required per zone
+    cpu_success_buffer = (1 - cpu_headroom_target(instance, buffers)) * 100
```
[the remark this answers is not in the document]
ayushisingh29: cpu_headroom_target will take care of instance headroom, background and desired buffer. In #721 I am only calculating derived buffer which will compute scale or preserve. Thats not counted in cpu_headroom_target.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+        new_cpu_utilization = current_cpu_utilization * scale
+        core_scale_up_factor = max(1.0, new_cpu_utilization / cpu_success_buffer)
+        return math.ceil(current_cores * core_scale_up_factor)
+    elif preserve:
+        return current_cores
+    else:
+        return int(current_cores * (current_cpu_utilization / cpu_success_buffer))
+
+
+def get_memory_with_buffer_gib(current_capacity, buffers):
```
ramsrivatsa, line 735: IMO, a buffer is just a placeholder. You should have 
`__get_disk_from_desire`
`_get_disk_from_current`

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -93,75 +156,51 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
...
+    needed_cores = math.ceil(capacity_requirement.cpu_cores.mid)
+    needed_disk = capacity_requirement.disk_gib.mid
+    needed_network_mbps = capacity_requirement.network_mbps.mid
 
-    needed_disk = math.ceil(
-        (1.0 / desires.data_shape.estimated_compression_ratio.mid)
-        * desires.data_shape.estimated_state_size_gib.mid
-        * copies_per_region,
-    )
+    if needed_disk == 0:
```
ramsrivatsa, line 178: Why would this return 0? We need guardrails for this to be non-zero.

ideally, the `get_disk_*` should be the ceiling and hence should be at least one. If in cases we get zero we might need to modify the internal functions for it to not return zero.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -93,75 +156,51 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
...
+            1, _get_disk_from_desires(desires, copies_per_region) // zones_per_region
+        )
+        # We want to compute memory based on the data size.
+        disk_used_gib = needed_disk / disk_buffer.ratio
 
     # Rough estimate of how many instances we would need just for the CPU
     # Note that this is a lower bound, we might end up with more.
-    rough_count = math.ceil(needed_cores / instance.cpu)
-
+    rough_count = math.ceil((needed_cores * zones_per_region) / instance.cpu)
```
ramsrivatsa, line 187: `estimated_cores_per_region`
this is more readable as a variable and more reflective of what we do in the previous steps.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -84,6 +141,12 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
     The input desires should be the **regional** desire, and this function will
     return the zonal capacity requirement
     """
+    disk_buffer = buffer_for_components(
```
ramsrivatsa, line 144: Move this lower to the place where we use it. Near line 178 so that people are aware of why this exists. Initially I was confused if you are repeating buffer allocation twice. 

After several times I understood that you want the computed buffer object to be used here for something else. Moving it closer to where it is used makes the code more readable.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
     return CapacityPlan(
         requirements=merged_requirements, candidate_clusters=merged_clusters
     )
+
+
+def derived_buffer_for_component(buffer, components):
```
ramsrivatsa, line 697:
```
def derived_buffer_for_component(buffer: Dict[str, Buffer], components: List[str]):
    scale = 0
    preserve = False

    if not buffer:
        return scale, preserve

    for bfr in buffer.values():
        if any(component in components for component in bfr.components):
            if bfr.intent == BufferIntent.scale:
                scale = max(scale, bfr.ratio)
            elif bfr.intent == BufferIntent.preserve:
                preserve = True

    return scale, preserve
```

1. Reducing nesting levels for more readability. 
2. s/i/component

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+                if bfr.intent == BufferIntent.scale:
+                    scale = max(scale, bfr.ratio)
+                if bfr.intent == BufferIntent.preserve:
+                    preserve = True
+
+    return scale, preserve
+
+
+def get_cores_with_buffer(current_capacity, buffers, instance):
+    # compute cores required per zone
```
ramsrivatsa, line 713: Please write in comments elaborately the example given here.
https://docs.google.com/document/d/1W6OwLlxrh5kLfOmOlmvJywfo4PmH1i6HUgEdL-9Lg4w/edit?tab=t.0

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+    if scale > 0:
+        # if the new cpu core is less than the current,
+        # then take no action and return the current cpu cores
+        new_cpu_utilization = current_cpu_utilization * scale
+        core_scale_up_factor = max(1.0, new_cpu_utilization / cpu_success_buffer)
+        return math.ceil(current_cores * core_scale_up_factor)
+    elif preserve:
+        return current_cores
+    else:
+        return int(current_cores * (current_cpu_utilization / cpu_success_buffer))
```
ramsrivatsa, line 732: You can avoid the else here. Make 721 return 1.0 for the case that comes here.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+
+
+def get_memory_with_buffer_gib(current_capacity, buffers):
+    # compute memory required per zone
+    current_memory_utilization = current_capacity.memory_utilization_gib.mid
+    current_ram_allocated = (
```
ramsrivatsa, line 738: s/current_ram_allocated/total_ram_allocated or zonal_ram_allocated

trying to be more explicit on context.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+    memory_buffer = buffer_for_components(
+        buffers=buffers, components=[BufferComponent.memory]
+    )
+
+    scale, preserve = derived_buffer_for_component(
+        buffers.derived, ["memory", "storage"]
+    )
+
+    if scale > 0:
+        # if the new required memory is less than the current,
```
ramsrivatsa, line 753: Ideally we need to reduce this. If this work is supposed to be planned to be executed in future, add a TODO here.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+        # then take no action and return the current ram
+        return max(
+            current_memory_utilization * scale * memory_buffer.ratio,
+            current_ram_allocated,
+        )
+    elif preserve:
+        return current_ram_allocated
+    else:
```
ramsrivatsa, line 761: Same in the previous case, the last else can be avoided.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+        * current_capacity.cluster_instance_count.mid
+    )
+
+    # These are the desired buffers
+    memory_buffer = buffer_for_components(
+        buffers=buffers, components=[BufferComponent.memory]
+    )
+
+    scale, preserve = derived_buffer_for_component(
+        buffers.derived, ["memory", "storage"]
```
ramsrivatsa, line 749: Could disk also be affected here. Can we make it extendable so that the buffer components from the tech itself. 

There could be a situation where EVCache memory buffer computation might have no impact on storage. However, for Cassandra it might have some impact. Hence, having the buffer components passed in for each tech might be a good idea.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+        # if the new required network is less than the current,
+        # then take no action and return the current bandwidth
+        return max(
+            current_network_utilization * scale * network_buffer.ratio,
+            current_network_allocated,
+        )
+    elif preserve:
+        return current_network_allocated
+    else:
+        return current_network_utilization * network_buffer.ratio
```
ramsrivatsa, line 792: same in the previous case.

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,157 @@ def merge_plan(
...
+        current_capacity.cluster_instance.drive.max_size_gib
+        * current_capacity.cluster_instance_count.mid
+    )
+
+    # These are the desired buffers
+    disk_buffer = buffer_for_components(
+        buffers=buffers, components=[BufferComponent.disk]
+    )
+
+    scale, preserve = derived_buffer_for_component(buffers.derived, ["storage", "disk"])
```
ramsrivatsa, line 808: Again same goes here too. There is a gray area between memory, storage and disk.Maybe this would be a good point to bring up on the expectations with a broader audience.

ramsrivatsa (changes requested):
```
diff --git a/service_capacity_modeling/models/common.py b/service_capacity_modeling/models/common.py
index 829959e..325b814 100644
--- a/service_capacity_modeling/models/common.py
+++ b/service_capacity_modeling/models/common.py
@@ -694,22 +694,24 @@ def merge_plan(
     )
 
 
-def derived_buffer_for_component(buffer, components):
+def derived_buffer_for_component(buffer: Dict[str, Buffer], components: List[str]):
     scale = 0
     preserve = False
 
-    if buffer is not None:
-        for bfr in buffer.values():
-            if any(i in components for i in bfr.components):
-                if bfr.intent == BufferIntent.scale:
-                    scale = max(scale, bfr.ratio)
-                if bfr.intent == BufferIntent.preserve:
-                    preserve = True
+    if not buffer:
+        return scale, preserve
+
+    for bfr in buffer.values():
+        if any(component in components for component in bfr.components):
+            if bfr.intent == BufferIntent.scale:
+                scale = max(scale, bfr.ratio)
+            elif bfr.intent == BufferIntent.preserve:
+                preserve = True
 
     return scale, preserve
 
 
-def get_cores_with_buffer(current_capacity, buffers, instance):
+def get_cores_with_buffer(current_capacity: CurrentClusterCapacity, buffers: Buffers, instance: Instance):
     # compute cores required per zone
     cpu_success_buffer = (1 - cpu_headroom_target(instance, buffers)) * 100
     current_cpu_utilization = current_capacity.cpu_utilization.mid
@@ -844,5 +846,8 @@ def zonal_requirements_from_current(
             network_mbps=certain_float(needed_network_mbps),
             reference_shape=current_capacity.cluster_instance,
         )
+    else:
+        raise ValueError(
+            f"Please check if current_cluster is populated correctly. "
+        )
 
-    return None
diff --git a/service_capacity_modeling/models/org/netflix/cassandra.py b/service_capacity_modeling/models/org/netflix/cassandra.py
index 3e5a8f6..6d51988 100644
--- a/service_capacity_modeling/models/org/netflix/cassandra.py
+++ b/service_capacity_modeling/models/org/netflix/cassandra.py
@@ -144,9 +144,7 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
     disk_buffer = buffer_for_components(
         buffers=desires.buffers, components=[BufferComponent.disk]
     )
-
     reference_shape = desires.reference_shape
-
     current_capacity = (
         None
         if desires.current_clusters is None
@@ -156,6 +154,8 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
             else desires.current_clusters.regional[0]
         )
     )
+
+    # If the cluster is not yet provisioned
     if current_capacity and desires.current_clusters is not None:
         capacity_requirement = zonal_requirements_from_current(
             desires.current_clusters, desires.buffers, instance, reference_shape
@@ -166,6 +166,7 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
         )
         disk_used_gib = current_capacity.disk_utilization_gib.mid * (disk_scale or 1)
     else:
+        # If the cluster is already provisioned
         capacity_requirement = _zonal_requirement_for_new_cluster(
             desires, instance, copies_per_region, zones_per_region
         )
@@ -193,6 +194,7 @@ def _estimate_cassandra_requirement(  # pylint: disable=too-many-positional-argu
         max(1, (disk_used_gib * zones_per_region) // rough_count)
     )
     rps_working_set = min(1.0, disk_rps / max_rps_to_disk)
+
     # If disk RPS will be smaller than our target because there are no
     # reads, we don't need to hold as much data in memory.
     # For c*, we can skip memory buffer and can just keep using the heap and write buffer calc

```

I had some super micro nits that I pasted here. Feel free to discard this patch. 

Also add comments wherever its appropriate.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -69,13 +72,67 @@ def _write_buffer_gib_zone(
...
+    # Do not add disk buffers now as memory calculation is done on the disk usage
+    return math.ceil(
+        (1.0 / desires.data_shape.estimated_compression_ratio.mid)
+        * desires.data_shape.estimated_state_size_gib.mid
+        * copies_per_region
+        * disk_buffer.ratio
+    )
+
+
+def _zonal_requirement_for_new_cluster(
```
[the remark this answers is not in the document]
ayushisingh29: Plan uncertain will call the **_estimate_cassandra_requirement** internally to get the desires.

service_capacity_modeling/models/common.py
ayushisingh29: The reason I added buffer because the final computed memory takes the required buffer into account(just not the usage). I will update the name to address your comment.

service_capacity_modeling/models/org/netflix/cassandra.py
ayushisingh29: If there is no disk_utilization_gib passed in the desires, then it will result in disk as 0. In that case we will use different params to compute the disk usage.

service_capacity_modeling/models/common.py
ayushisingh29: This is for the cases where we need to perform the right sizing. We will only perform right sizing if there is no scale and preserve.

ramsrivatsa (approved):

service_capacity_modeling/models/common.py
```diff
@@ -689,3 +692,198 @@ def merge_plan(
...
+        cluster_instance = shapes.instance(current_capacity.cluster_instance_name)
+    else:
+        cluster_instance = current_capacity.cluster_instance
+
+    current_cores = cluster_instance.cpu * current_capacity.cluster_instance_count.mid
+
+    scale, preserve = derived_buffer_for_component(buffers.derived, ["compute", "cpu"])
+    # Scale and preserve for the same component should not be passed together.
+    # If user passes it, then scale will be preferred over preserve.
+    if scale > 0:
```
nickmahilani, line 731: shouldn't scale be > 1 here? If scale is 0.5, then we effectively want to preserve?

ayushisingh29: Here, `core_scale_up_factor = max(1.0, new_cpu_utilization / cpu_success_buffer)` we are taking the max of 1 or  new cpu factor. If new CPU factor is low then we will keep the existing cores.

PR #128, opened 2025-03-18
Add c5, m4, m6a, and m6i specs

jasonk000:
As requested by @nbehnam.

### DIFF
--- a/service_capacity_modeling/hardware/profiles/shapes/aws/auto_m6i.json
+++ b/service_capacity_modeling/hardware/profiles/shapes/aws/auto_m6i.json
@@ -5,7 +5,7 @@
       "cpu": 2,
       "cpu_cores": 1,
       "cpu_ghz": 3.5,
-      "cpu_ipc_scale": 1.0,
+      "cpu_ipc_scale": 1.15,
       "ram_gib": 7.63,
       "net_mbps": 781.0,
       "drive": null
@@ -15,7 +15,7 @@
       "cpu": 4,
       "cpu_cores": 2,
       "cpu_ghz": 3.5,
-      "cpu_ipc_scale": 1.0,
+      "cpu_ipc_scale": 1.15,
       "ram_gib": 15.26,
       "net_mbps": 1562.0,
       "drive": null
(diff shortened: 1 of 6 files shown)


service_capacity_modeling/hardware/profiles/shapes/aws/auto_m6i.json
```diff
@@ -15,7 +15,7 @@
       "cpu": 4,
       "cpu_cores": 2,
       "cpu_ghz": 3.5,
-      "cpu_ipc_scale": 1.0,
+      "cpu_ipc_scale": 1.15,
```
jolynch, line 18: Oh interesting, I thought all of the 15% uplift from m5 to m6i was just the 3.1 -> 3.5. There was a 30% lift from 5th to 6th?

jasonk000: Dependent on workload, yes. Ice Lake has a significant IPC bump for many workloads (especially crypto and AVX-512). The turbo curve also shifted; the sustained all-core turbo on m6i is quite flat with sustained all-core at 3.5, which is a difference from m5.

jolynch: I'm inclined to leave these ipc_scale numbers alone unless we change all of the (m|r|c)6i and bump the m7i at the same time or base them all on an IPC benchmark.

We really just added that scale factor to account for m7a cores being significantly faster and to make it so m7i (3.2G clock) didn't come out as _slower_ than m6i (3.5G clock). With this change the 6th gen is now faster again (3.5 * 1.15 = 4.025 effective versus 7th gen 3.2 * 1.25 = 4 effective).

We could also split this into two parts, one that adds the shapes without making an IPC claim and the other that refactors _all_ affected shapes in the 6th gen.

jolynch:
Rebased to make this clean again after #135, might be easier to drop the 6th gen IPC changes out of this PR and do a follow-up which modifies _all_ of the 6th gen? Up to you

PR #133, opened 2025-04-18
Integrate kafka capacity planner with buffers

ramsrivatsa:


service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -506,6 +496,27 @@ def default_desires(user_desires, extra_model_arguments: Dict[str, Any]):
...
         state_gib = (write_bytes.mid * retention_secs) / GIB_IN_BYTES
 
+        # By supplying these buffers we can deconstruct observed utilization into
+        # load versus buffer.
+        buffers = Buffers(
+            default=Buffer(ratio=1.5),
+            desired={
+                # Amount of compute buffer that we need to reserve in addition to
+                # cpu_headroom_target that is reserved on a per instance basis
+                "compute": Buffer(ratio=1.5, components=[BufferComponent.compute]),
```
nickmahilani, line 506: ratio=2.5 here for tier0 and tier1 (to target 40% CPU utilization max on instance). It can be lower(1.5) for tier=2 and tier=3

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -506,6 +496,27 @@ def default_desires(user_desires, extra_model_arguments: Dict[str, Any]):
...
+                # This makes sure we use only 25% of the available storage
+                "storage": Buffer(ratio=4.0, components=[BufferComponent.storage]),
```
nickmahilani, line 508: this can be 2.5X for storage (we are ok using upto 40% of disk)

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -506,6 +496,27 @@ def default_desires(user_desires, extra_model_arguments: Dict[str, Any]):
...
+                "background": Buffer(
+                    ratio=2.0,
+                    components=[
+                        BufferComponent.cpu,
```
nickmahilani, line 512: how is `background cpu buffer` different from the `compute buffer` above?

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -100,39 +104,34 @@ def _estimate_kafka_requirement(  # pylint: disable=too-many-positional-argument
...
+            reference_shape=desires.reference_shape,
         )
-        logger.debug("kafka needed cores: %s", needed_cores)
-        # Normalize those cores to the target shape
-        reference_shape = current_zonal_cluster.cluster_instance
-        needed_cores = normalize_cores(
-            core_count=needed_cores,
-            target_shape=instance,
-            reference_shape=reference_shape,
+        needed_cores = int(capacity_requirement.cpu_cores.mid)
```
nickmahilani, line 128: just using mid cpu cores for needed cores felt not very accurate, hence I had the logic that you have removed. What is rationale for using the cpu.mid here instead?

PR #135, opened 2025-04-21
LSCALE-157: Add m6a instance family

ribhavjain:
Add m6a instance family

### DIFF
--- a/service_capacity_modeling/hardware/profiles/shapes/aws/auto_m6a.json
+++ b/service_capacity_modeling/hardware/profiles/shapes/aws/auto_m6a.json
@@ -1,0 +1,104 @@
+{
+  "instances": {
+    "m6a.large": {
+      "name": "m6a.large",
+      "cpu": 2,
+      "cpu_cores": 1,
+      "cpu_ghz": 3.6,
+      "cpu_ipc_scale": 1.0,
+      "ram_gib": 7.63,
+      "net_mbps": 781.0,
+      "drive": null
+    },
+    "m6a.xlarge": {
+      "name": "m6a.xlarge",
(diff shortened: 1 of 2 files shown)


shy-1234:
There is an assertion fail in test_kafka_high_throughput on the expected_disk value range. The reason is because after adding m6a family into the shapes, the least regret plan come out of capacity modeler is now suggesting "m6a.4xlarge" instead of the previous "m7i.4xlarge" and that causes the requirements.disk_gib.mid value changed.

jolynch:
That test is fwict asking for RF=2, across three zones at 2.8GiB/s write throughput, which by my math is minimum 20 TiB / zone:

```
You have: 2 * ((2.8 * 1024 * 1024 * 1024 bytes / second) / 3)
You want: gibibytes / 3 hours
        * 20160
        / 4.9603175e-05
```

Should be safe to reduce the lower bound of the test to 22 TiB from 24 so that assertion passes imo, but @nickmahilani let me know if you feel otherwise.

jolynch:
Interesting, just noticed that Jason had a PR open over in #128 that also introduced the m6a but with a 1.2x ipc factor and m6i with 1.15.

I'm skeptical that the 6th gen intel achieved faster IPC (3.5 * 1.15) than the m7i (3.2 * 1.25 = 4) ... which I would think is what the m6a@1.2x claim is saying. Let's merge this so you get network values then we can either merge his on top, or adjust up the m7i while we are moving the ipc values around for m6i.

Curious if perf is going to go through and replace all of these guesses with actual relative avg cpu perf.

ribhavjain:
Sounds good, thanks for having a look

PR #137, opened 2025-05-07
Missing instance families for AWS

jasonk000:


service_capacity_modeling/tools/instance_families.py
```diff
@@ -0,0 +1,65 @@
...
+
+HSW_IPC = 0.85
+SKX_IPC = 1.0
+ICX_IPC = SKX_IPC * 1.15 # harshad says 1.15, chatgpt says 1.18
+SPR_IPC = ICX_IPC * 1.12 # harshad says 1.12, chatgpt says 1.19
+ROME_IPC = 1.03
+MILAN_IPC = SKX_IPC * 1.15
+GENOA_IPC = MILAN_IPC * 1.13
+
+instance_families = {
```
jasonk000, line 15: this and below is where the magic happens

tests/test_hardware_shapes.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def get_instance_families() -> Dict[str, List[Tuple[str, str]]]:
+    """Extract all instance families from available instances."""
+    region = shapes.region("us-east-1")
+    families: Dict[str, List[Tuple[str, str]]] = {}
+    for instance_name in region.instances:
+        match = re.match(r"([a-z]+\d+[a-z]*)\.(\d+xl|[a-z]+)", instance_name)
+        if match:
+            family, size = match.groups()
```
jolynch, line 22: nit: there are already helpers on the Instance objects themselves to get the family and shape from the instance name - this is a little fragile to if we change the names in the keys (which are unstructured)

tests/netflix/test_cassandra_uncertain.py
```diff
@@ -108,7 +108,7 @@ def test_increasing_qps_simple():
         )
 
     # We should generally want cheap CPUs
-    assert all(r[0][0] in ("r", "m", "i") for r in result)
+    assert all(r[0][0] in ("r", "m", "i", "c") for r in result)
```
jolynch, line 111: I don't think we need to do it in this PR, but I think this kind of refactor exposes the fragility of direct family comparisons, it seems that we were missing a lot of the `c` shapes which were actually fine most of the time. @abersnaze let's think how we can refactor some of the "shape" check tests to be more robust to evolving hardware - I started down the path of cost / cpu_power / "at least as good as X cluster" / "just check resources" but clearly we have a lot of work to do

tests/test_hardware_shapes.py
```diff
@@ -0,0 +1,190 @@
...
+            continue
+
+        # Calculate memory per CPU core for all instances in this family
+        mem_to_cpu_ratios: List[Tuple[str, float]] = []
+        for instance_name, _ in instances:
+            instance = region.instances[instance_name]
+            ratio = instance.ram_gib / instance.cpu
+            mem_to_cpu_ratios.append((instance_name, ratio))
+
+        # All ratios should be approximately the same within a family
```
jolynch, line 150: Nice

service_capacity_modeling/tools/instance_families.py
```diff
@@ -0,0 +1,265 @@
...
+        "io_latency_curve": "6th-gen-ssd",
+        "cpu_ipc_scale": ICX_IPC,
+        "cpu_turbo_single_ghz": 3.5,
+        "cpu_turbo_all_ghz": 3.5,
+    },
+    "r7a": {
+        "xl_iops": None,
+        "io_latency_curve": None,
+        "cpu_ipc_scale": GENOA_IPC,
+        "cpu_turbo_single_ghz": 3.7,
```
jolynch, line 253: How do the turbo affect the calculations or are you just future proofing? Definitely leave it to y'all on if this is how you want to maintain it, definitely agreed this plus auto_shape is probably easier than before, but I am curoius if this needs shape level granularity (one thing I liked about the auto_shape was we more expressed those opinions for a generation/architecture pair rather than per shape).

service_capacity_modeling/tools/generate_missing.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    for family, cmd in commands.items():
+        if debug:
+            print(f"\nWould run for {family}:\n{' '.join(cmd)}")
+
+        if execute:
+            print(f"\nGenerating {family}...")
+            try:
+                result = subprocess.run(cmd, capture_output=True, text=True, check=True)
```
jolynch, line 74: This is fine, but could also just import and call the functions you need (e.g. call with args).

service_capacity_modeling/tools/instance_families.py
```diff
@@ -0,0 +1,265 @@
...
+
+HSW_IPC = 0.85
+SKX_IPC = 1.0
+ICX_IPC = SKX_IPC * 1.15  # harshad says 1.15, chatgpt says 1.18
+SPR_IPC = ICX_IPC * 1.12  # harshad says 1.12, chatgpt says 1.19
+ROME_IPC = 1.03
+MILAN_IPC = SKX_IPC * 1.15
+GENOA_IPC = MILAN_IPC * 1.13
+
+INSTANCE_TYPES = {
```
jolynch, line 11: I think this belongs in hardware, but we can leave it in tools if you prefer (or when we kinda figure out the shape of this data we like / can automate we could shift it then).

service_capacity_modeling/hardware/profiles/shapes/aws/auto_c7i.json
```diff
@@ -5,7 +5,7 @@
       "cpu": 2,
       "cpu_cores": 1,
       "cpu_ghz": 3.2,
-      "cpu_ipc_scale": 1.25,
+      "cpu_ipc_scale": 1.288,
```
jolynch, line 8: So precise :laughing:

jolynch (approved):
Looks good to me, a few suggestions but nothing blocking, we can tackle those followups later (or not at all).

tests/netflix/test_cassandra_uncertain.py
abersnaze: Maybe the script that generates family shapes can pre-compute the likely lowest-cost prefix that the tests should expect to see.

tests/test_hardware_shapes.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def get_generation_groups() -> Dict[str, List[Tuple[str, int]]]:
+    """Group instance families by their base type and sort by generation."""
+    families = get_instance_families()
+    generation_groups: Dict[str, List[Tuple[str, int]]] = {}
+
+    for family in families:
+        # Extract the base type (c, m, r) and generation number
+        match = re.match(r"([a-z]+)(\d+)([a-z]*)", family)
```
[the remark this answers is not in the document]
jasonk000: Almost, but not quite. The logic in auto_shape parses the base type and generation, but is agnostic to variants (eg: c5 vs c5d vs c5dn).

re: i4i, the test only asserts that within a "base" type and "variant" that the performance increases gen/on/gen, such as i7 > i4 > i3, it does not compare that i4 == m4 or anything similar.

https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/main/service_capacity_modeling/tools/auto_shape.py#L267-L279

PR #140, opened 2025-05-22
Run linters before end-to-end test suite for faster build result

rayiniv-nflx:
Currently, the linter runs as the last step in the build. This results in builds failing for lint issues (if not checked locally) after 30-40 mins after the end-to-end test suite finishes running. Instead we can move it up before the tests so it can fail faster and PR submitter can fix sooner rather than waiting another 30 mins.


jolynch:
Nice! Thank you!

jolynch (commented):
:ship: lg2m

PR #141, opened 2025-05-27
Restrict Instance Type Param + Change Tier Target CPU Bounds

rayiniv-nflx:
Adding an option to restrict the instance types returned by the capacity planner to only use the same instance family as the current cluster if it exists. By default this will be True as that is the majority case. For new cluster provisioning, this check will bypass since there is no current cluster value anyways.

Also changing target CPU Bounds for tiers as below (open to suggestions for different values here):
- tier0: 40% cpu
- tier1: 45% cpu
- others: 50% cpu

### DIFF
--- a/service_capacity_modeling/models/org/netflix/kafka.py
+++ b/service_capacity_modeling/models/org/netflix/kafka.py
@@ -18,6 +18,7 @@ from service_capacity_modeling.interface import certain_float
 from service_capacity_modeling.interface import certain_int
 from service_capacity_modeling.interface import Clusters
 from service_capacity_modeling.interface import Consistency
+from service_capacity_modeling.interface import CurrentZoneClusterCapacity
 from service_capacity_modeling.interface import DataShape
 from service_capacity_modeling.interface import Drive
 from service_capacity_modeling.interface import GIB_IN_BYTES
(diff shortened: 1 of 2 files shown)


service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -47,13 +48,30 @@ def target_cpu_utilization(tier: int) -> float:
...
     """
-    if tier == 0:
-        return 0.25
-    if tier == 1:
-        return 0.30
-    if tier == 2:
-        return 0.35
-    return 0.40
+    if tier in (0, 1):
+        return 0.40
```
nickmahilani, line 52: I guess this will evolve with buffer integration. We have to think about what the incoming metric is for current cluster cpu util (think it is max of avg cpu util on any broker) and how to translate that to have sufficient success buffer when computing needed_cores

rayiniv-nflx: Yea, the current cpu util metric is using the atlas metric `kafka-cluster-capacity-cpu-average` which is what we were using when we do the spreadsheets as well. This indeed should be a max of avg cpu per broker as you mentioned. With the buffer logic added, I think this method may not be as useful as it should be encapsulated in the desired buffer ratios (2.5x buffer ~ 40% cpu util peak).

PR #142, opened 2025-05-28
Change buffers implentation to assume per node utilization

akashdeepgoel:


service_capacity_modeling/models/common.py
```diff
@@ -745,7 +745,10 @@ def get_memory_from_current_capacity(
     current_capacity: CurrentClusterCapacity, buffers: Buffers
 ):
     # compute memory required per zone
-    current_memory_utilization = current_capacity.memory_utilization_gib.mid
+    current_memory_utilization = (
+        current_capacity.memory_utilization_gib.mid
+        * current_capacity.cluster_instance_count.mid
```
jolynch, line 750: Line 758 seems to be already converting to a memory requirement, why do we need to double multiply?  This input data is supposed to represent the range of utilization over the cluster, e.g. capture the statistical distribution of instance utilization (picking min versus high has consequences, or you can do what the stateless internal model does and pick the p95).

Let's add tests of this function (and the cpu/network/disk ones) as well since this kind of change _should_ have broken unit tests that assert this.

jolynch (changes requested):
I am not sure this is consistent with these functions, and ideally would like to see test coverage of these functions @ayushisingh29 that would catch this kind of change (and signal that the dev is changing the meaning of these common functions).

PR #143, opened 2025-06-02
Adding buffers to kafka cap model

rayiniv-nflx:
Replaces the current cluster scaling logic in kafka cap model with the `zonal_requirements_from_current` common function.  Removing other unnecessary logic associated with the older calculation method.

This PR also depends on the logic in [PR#142](https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/142) which fixes the common buffer implementation to use per node metrics.

### DIFF
--- a/tests/netflix/test_kafka.py
+++ b/tests/netflix/test_kafka.py
@@ -376,9 +376,9 @@ def test_plan_certain_data_shape():
         ),
         memory_utilization_gib=Interval(low=0, mid=0, high=0, confidence=1),
         network_utilization_mbps=Interval(
-            low=4580.919447446355,
-            mid=19451.59814477331,
-            high=42963.441154527085,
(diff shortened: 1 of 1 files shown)


tests/netflix/test_kafka.py
```diff
@@ -546,3 +556,132 @@ def test_plan_certain_data_shape_same_instance_type():
 
     for lr in cap_plan:
         print(lr.candidate_clusters.zonal[0])
+
+
+def test_plan_certain_data_shape_buffers():
```
nickmahilani, line 561: naming nit: test_scale_up_using_buffers

nickmahilani (approved):

ramsrivatsa:
Thanks @nickmahilani for reviewing this. This PR got underwater. My apologies.

PR #144, opened 2025-06-05
Allowing multiple capacity plans for the same family

rayiniv-nflx:
Currently, the capacity planner reduces the possible plans to allow at most 1 plan per instance family. This may not always be ideal in cases where the capacity planner may recommend a lower instance type with a higher count in certain cases due to disk requirements. i.e. `20x r5.2xl instances with 1TB disk` vs. `10x r5.4xl instances with 2TB disk` where the higher count ends up costing less and is the top recommendation by the planner but it is more inconvenient from a scaling perspective as we would rather stay closer to the current instance count and vertically scale instead. In such cases, the operator could benefit from examining multiple options before making a decision.

This PR aims to allow multiple results for the same family to help the above issue. By default this will be 1 so the behavior is the same for anyone not using this feature (backwards compatible). Central scaling workflow may not use this feature as it likely will require one final recommendation.

Ideally the planner itself should converge to the most ideal plan in general but this can also help the operator debug/example the planner results more easily.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -577,7 +577,10 @@ class CapacityPlanner:
         plans.sort(key=lambda p: (p.rank, p.candidate_clusters.total_annual_cost))
 
         num_results = num_results or self._default_num_results
-        return reduce_by_family(plans)[:num_results]
+        max_per_family = extra_model_arguments.get("max_results_per_family", 1)
+        return reduce_by_family(plans, max_results_per_family=max_per_family)[
+            :num_results
+        ]
 
     # Calculates the minimum cpu, memory, and network requirements based on desires.
     def _per_instance_requirements(self, desires) -> Tuple[int, float]:
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -511,6 +511,7 @@ class CapacityPlanner:
         num_results: Optional[int] = None,
         num_regions: int = 3,
         extra_model_arguments: Optional[Dict[str, Any]] = None,
+        max_results_per_family: int = 1,
     ) -> Sequence[CapacityPlan]:
         if model_name not in self._models:
             raise ValueError(
@@ -538,6 +539,7 @@ class CapacityPlanner:
                 lifecycles=lifecycles,
                 instance_families=instance_families,
                 drives=drives,
+                max_results_per_family=max_results_per_family,
             )
             if sub_plan:
                 results.append(sub_plan)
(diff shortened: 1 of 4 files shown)


service_capacity_modeling/capacity_planner.py
```diff
@@ -577,7 +577,10 @@ def _plan_certain(  # pylint: disable=too-many-positional-arguments
         plans.sort(key=lambda p: (p.rank, p.candidate_clusters.total_annual_cost))
 
         num_results = num_results or self._default_num_results
-        return reduce_by_family(plans)[:num_results]
+        max_per_family = extra_model_arguments.get("max_results_per_family", 1)
```
jolynch, line 580: If we do end up lifting to planner, we'd probably codify in the api as a kwarg, since it's not a parameter of the workload (the desires) but rather the plan. So like how num_results or lifecycles/instance_familes etc work above

tests/test_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+from service_capacity_modeling.interface import CapacityPlan
+from service_capacity_modeling.interface import Clusters
+from service_capacity_modeling.interface import Instance
+from service_capacity_modeling.interface import Requirements
+from service_capacity_modeling.interface import ZoneClusterCapacity
+from service_capacity_modeling.models.utils import reduce_by_family
+
+
+class TestUtils(unittest.TestCase):
```
jolynch, line 14: Can you use [pytest](https://docs.pytest.org/en/stable/) instead of unittest to match the rest of the suite? Most python devs don't expect java style tests like this. So nix the class, call your test methods "test_..." and then just use plain asserts not class methods.

rayiniv-nflx: Sounds good. Updated test class removing unittest in favor of pytest

service_capacity_modeling/models/utils.py
```diff
@@ -31,11 +38,20 @@ def reduce_by_family(plans: Iterable[CapacityPlan]) -> List[CapacityPlan]:
...
             )
 
-        if not (zonal_type in zonal_families and regional_type in regional_families):
+        # Count how many of each family combination we've seen
+        zonal_count = zonal_families.get(zonal_type, 0)
+        regional_count = regional_families.get(regional_type, 0)
+
+        # Add the plan if we haven't reached the maximum for either family type
+        if (
+            zonal_count < max_results_per_family
```
ayushisingh29, line 47: If max_results_per_family is 3 then according to this we can have 3 zonal count and 3 regional count. Is it expected or will it  be 3 for zonal and regional combined?

ayushisingh29 (approved):

PR #147, opened 2025-06-13
Add validation when required clusters is below minimum threshold (2) for critical tiers (0,1)

homatthew:
Occasionally, tier 1 or 0 clusters can be provisioned as tier < 1 initially with only 1 node per ASG. This leads to issues with the capacity model because we've restricted horizontal scaling but the min_count for tier 1 clusters is 2 nodes per asg.

The `min_count` variable is used by the uncertain and certain model to enforce this restriction. But there are times where clusters change tier-ing and this model is no longer able to process them.

There proposed solution is:
* Create a pre condition to loudly fail with an input of required_cluster_size is set

NOTE: This code path is shared by both the uncertain and certain models


service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -556,6 +557,33 @@ class NflxCassandraArguments(BaseModel):
 
 
 class NflxCassandraCapacityModel(CapacityModel):
+    @staticmethod
+    def get_required_cluster_size(tier, extra_model_arguments):
```
ayushisingh29, line 561: Can we add tests for this?

tests/test_cassandra.py
```diff
@@ -0,0 +1,51 @@
...
+            ValueError,
+        ),
+    ],
+)
+def test_get_required_cluster_size_exceptions(
+    tier, extra_model_arguments, expected_exception
+):
+    with pytest.raises(expected_exception):
+        NflxCassandraCapacityModel.get_required_cluster_size(
+            tier, extra_model_arguments
```
ayushisingh29, line 50: assert the test result here.

homatthew: Naive question: Why are we asserting a result if we expect no result when there's an exception?

homatthew: Oh maybe this is a misunderstanding. These are 2 separate tests. They both are separate tests that end up computing the following tests:

image

PR #148, opened 2025-06-13
Add ability to force regenerate the instance shapes in `generate_missing.py`

homatthew:
Not sure if this is the correct change. I noticed this discrepancy when understanding the context in this thread
- https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/146#discussion_r2145642132


jasonk000 (changes requested):
@homatthew Change looks great, however, we should separate the inclusion of `--force` and the update of the JSONs into separate PRs.

jasonk000 (approved):

PR #151, opened 2025-06-21
Fixing non-ebs cluster planning

rayiniv-nflx:
Kafka planner was not accounting for the local disk size for non-ebs clusters compared to attached disks for ebs clusters. This PR adjusts the validation in the Kafka specific planner to check the correct value before returning a plan.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/kafka.py
+++ b/service_capacity_modeling/models/org/netflix/kafka.py
@@ -246,7 +246,7 @@ def _estimate_kafka_cluster_zonal(  # noqa: C901
     require_attached_disks: bool = False,
     required_zone_size: Optional[int] = None,
     max_regional_size: int = 150,
-    max_local_disk_gib: int = 1024 * 16,  # i3en.6xl has 15TB disk
+    max_local_disk_gib: int = 1024 * 8,
     min_instance_cpu: int = 2,
     min_instance_memory_gib: int = 12,
     require_same_instance_family: bool = True,
@@ -494,7 +494,7 @@ class NflxKafkaCapacityModel(CapacityModel):
         max_regional_size: int = extra_model_arguments.get("max_regional_size", 150)
         # Very large nodes are hard to cache warm
         max_local_disk_gib: int = extra_model_arguments.get(
-            "max_local_disk_gib", 1024 * 16  # i3en.6xlarge can use up to 15TB
+            "max_local_disk_gib", 1024 * 8
         )
         min_instance_cpu: int = extra_model_arguments.get("min_instance_cpu", 2)
         min_instance_memory_gib: int = extra_model_arguments.get(
--- a/service_capacity_modeling/models/org/netflix/kafka.py
+++ b/service_capacity_modeling/models/org/netflix/kafka.py
@@ -247,7 +247,7 @@ def _estimate_kafka_cluster_zonal(  # noqa: C901
     require_attached_disks: bool = False,
     required_zone_size: Optional[int] = None,
     max_regional_size: int = 150,
-    max_local_disk_gib: int = 1024 * 8,
+    max_local_disk_gib: int = 1024 * 5,
     min_instance_cpu: int = 2,
     min_instance_memory_gib: int = 12,
     require_same_instance_family: bool = True,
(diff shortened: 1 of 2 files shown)


tests/netflix/test_kafka.py
```diff
@@ -859,3 +860,157 @@ def test_scale_up_using_buffers_high_disk_change_instance_count():
 
     for lr in cap_plan:
         print(lr.candidate_clusters.zonal[0])
```
ramsrivatsa, line 862: Remove prints.

tests/netflix/test_kafka.py
```diff
@@ -859,3 +860,157 @@ def test_scale_up_using_buffers_high_disk_change_instance_count():
...
+        * buffer_ratio
+        * scale_ratio
+    )
+    assert (
+        lr_clusters[0].instance.drive.size_gib * lr_clusters[0].count
+        >= minimum_provisioned_disk
+    )
+
+    for lr in cap_plan:
+        print(lr.candidate_clusters.zonal[0])
```
ramsrivatsa, line 1016: Remove prints.

ramsrivatsa (changes requested):
Fix comment and tests.

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -492,7 +494,7 @@ def capacity_plan(
         max_regional_size: int = extra_model_arguments.get("max_regional_size", 150)
         # Very large nodes are hard to cache warm
         max_local_disk_gib: int = extra_model_arguments.get(
-            "max_local_disk_gib", 1024 * 5
+            "max_local_disk_gib", 1024 * 16  # i3en.6xlarge can use up to 15TB
```
nickmahilani, line 497: am not sure if we want to have 16 TB of state on local disk, this is not a viable candidate instance type for us given the high cost to replace the instance

rayiniv-nflx: Makes sense. I have reduced this to 8TB. That being said, the model may still select i3en.6xl but it will limit the disk usage to at most 8TB on each instance. I have played around with lower values and we do not get any results for many of the large fronting Datamesh clusters since they are almost at the 150 broker limit that the model enforces as well. Having too low of a limit here would require breaking the 150 broker limit.
In general, these clusters are reaching their upper limit that we can manage scaling for so they likely need to be broken down further into smaller clusters. We can follow up on that separately.

nickmahilani: yeah we have to move topics out to other clusters if we are already at max cluster capacity. This should still be 5 TB IMHO

rayiniv-nflx: Yea agreed. We have a few EBS clusters at 7TB so felt 8 was fine here for now since these are easier to scale down as well after as well. The downside being the expensive replication/time cost if a node goes down. 
I think we probably do not have enough time to move topics on the Datamesh side before the next event so thinking we can keep this at 8 for now in order to get recommendations for the fronting clusters and reduce it to 5 after coordinating with Datamesh team to reduce disk usage on these clusters.
What do you think?

rayiniv-nflx: We could preserve the disk as well but I think disk is the main constraining factor for these clusters. If we do not scale the disk at all (preserve) we would be pretty underscaled for the event. Another option may be to use a lower desired buffer potentially but I am not sure if we want to run at a higher utilization before this event.

ramsrivatsa (approved):

service_capacity_modeling/models/org/netflix/kafka.py
rayiniv-nflx: Discussed offline. We will keep the limit as is at 5TB as this is the max disk usage we want on any single node. We should move the data off the cluster to a different one if we surpass this limit.

service_capacity_modeling/models/org/netflix/kafka.py
```diff
@@ -360,11 +361,13 @@ def _estimate_kafka_cluster_zonal(
...
-            if max_attached_disk_gib is not None
-            else drive.max_size_gib / 3
-        )  # Max allowed disk in `compute_stateful_zone`
+
+        # Max allowed disk size in `compute_stateful_zone`
+        max_size = drive.max_size_gib / 3
+        if instance.drive is not None and instance.drive.size_gib > 0:
+            max_size = min(max_local_disk_gib, instance.drive.size_gib)
+        elif max_attached_disk_gib is not None:
+            max_size = max_attached_disk_gib
```
ramsrivatsa, line 370: move line 366 to else:
        max_size = drive.max_size_gib / 3

ramsrivatsa (approved):

PR #152, opened 2025-06-23
add ebs support for Cassandra

ayushisingh29:


tests/netflix/test_cassandra_uncertain.py
```diff
@@ -150,72 +154,61 @@ def test_worn_dataset():
...
+        assert has_attached_storage(lr_cluster)
+        assert lr_cluster.attached_drives[0].name == "gp3"
+        assert_minimum_storage_gib(lr_cluster, 102_400)
+        # gp3 should not provision massive drives, prefer to upcolor
+        assert get_drive_size_gib(lr_cluster) <= 8 * 1024
+
+        # (matthewho) Why is this r6a.2xlarge??? The above test has the
+        # same desires and is m7a.xlarge. The only meaningful difference is
+        # the required_attached_disks. Also, the resolved disk size is larger
+        assert_similar_compute(
```
homatthew: This output is strange. The desires are the same as the previou stest but somehow the output is r6a.2xlarge (as opposed to r7a.xlarge).

It's hard to walk through the code in the uncertain code to understand why the outputs are different at all. I timeboxed this change short-term but would love some extra eyes to understand this behvaior better.

tests/netflix/test_cassandra.py
```diff
@@ -81,13 +84,8 @@ def test_capacity_small_fast(self):
...
             # Even though it's a small dataset we need IOs so should end up
-            # with lots of ebs_gp2 to handle the read IOs
-            if small_result.attached_drives:
-                assert (
-                    small_result.count
-                    * sum(d.size_gib for d in small_result.attached_drives)
-                    > 1000
-                )
+            # with lots of storage to handle the read IOs
+            assert_minimum_storage_gib(small_result, 1000)
```
ramsrivatsa: Assert storage is less than. 
I know that the intent of this test is that with high QPS and small data size we need the planner to show minimum storage. Write that in comments. 

your function should still be named assert storage is less than. 


very nice refactor.

tests/netflix/test_cassandra.py
```diff
@@ -101,23 +99,12 @@ def test_capacity_high_writes(self):
         )[0]
         high_writes_result = cap_plan.candidate_clusters.zonal[0]
         assert high_writes_result.instance.family.startswith("c")
-        assert high_writes_result.count > 4
+        assert high_writes_result.count >= 2
```
ramsrivatsa: Why did this go down by half?

homatthew: However, the idea that we can handle 100k writes using c7a.4xlarge using:

```
32 cores (c7a.4xlarge)
GP3 ebs with 13.4k read IOPS and 200 read IOPS
```

The question would be, why can't we use an equivalent m6id.4xlarge or c6id.4xlage? 


The answer is you can, but it will be much more exensive (12.5k) and (10.2k) respsectively as opposed to 7.15k. 

The cheapest non ebs configuration would be the 4x c6id.4xlarge.

homatthew: Upon further thought, I think the right thing to do is to raise the "high write" definition to match the generational improvements.  Here's some examples with 500k WPS whihc seems more suitable now for the amount of compute required

```
00 = {dict} {'annual_cost': 23010.64, 'count': 8, 'cpu': 16, 'instance': 'c7a.4xlarge'}
01 = {dict} {'annual_cost': 25644.0, 'count': 4, 'cpu': 48, 'instance': 'c5a.12xlarge'}
02 = {dict} {'annual_cost': 25948.0, 'count': 4, 'cpu': 48, 'instance': 'c6a.12xlarge'}
03 = {dict} {'annual_cost': 26618.64, 'count': 8, 'cpu': 16, 'instance': 'm7a.4xlarge'}
04 = {dict} {'annual_cost': 27366.68, 'count': 4, 'cpu': 48, 'instance': 'c5.12xlarge'}
05 = {dict} {'annual_cost': 28686.68, 'count': 4, 'cpu': 48, 'instance': 'c6i.12xlarge'}
06 = {dict} {'annual_cost': 29506.68, 'count': 4, 'cpu': 48, 'instance': 'c5d.12xlarge'}
07 = {dict} {'annual_cost': 29981.32, 'count': 4, 'cpu': 48, 'instance': 'm6a.12xlarge'}
```

PR #157, opened 2025-07-22
Create Cassandra clusters of non power of 2

homatthew:
C* supports arbitrary node counts but the capacity planner forces them to be a power of 2.  In general, it's easiest to only support doubling due to operations. 

A longer term vision is for the non power of 2 restriction to be removed but the token algorithm would need to be revised and there are edge cases with off by 1 token issues that will make maintenance tasks take longer


homatthew:
Adding the tests in #168 since the logic for handling powers of 2 was modified and updated in that PR.

PR #168, opened 2025-08-14
`scale_up` and `scale_down` Buffer Interface

homatthew:


service_capacity_modeling/models/common.py
```diff
@@ -713,37 +791,120 @@ def derived_buffer_for_component(buffer: Dict[str, Buffer], components: List[str
...
+            )
+        )
+
+        # Calculate the allocated amount using an adjusted disk size because
+        # sometimes the capacity planner provisions a cluster where each node
+        # has a large disk (e.g. 15TB for i3en.6xlarge). But the cluster is
+        # allocated assuming that we wouldn't want to use the full disk size
+        # (e.g. 5TB). So now that we are calculating the existing capacity, we
+        # also need to take into account the max node_density and not just
+        # use the full instance disk size as-is
```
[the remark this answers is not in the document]
jolynch: For all of these intents we're locking the cluster size anyways on technologies that only support vertical scale, so the density can't exceed reasonable amounts. Shapes that "over-provision" will just be expensive and therefore not chosen. I also am skeptical of a requirements based shape indication like this.

tests/test_common.py
```diff
@@ -535,39 +513,24 @@ def test_get_disk_with_buffer_scale_down():
...
         current_capacity=current_cluster_copy.zonal[0],
         buffers=buffers_copy,
-        instance_candidate=default_reference_shape,
     ).disk_gib
     expected_buffer = 4  # disk buffer
     assert needed_disk <= disk_utilization_gib.mid * cluster_size.mid * expected_buffer
     assert needed_disk == 14000 * cluster_size.mid
     assert needed_disk == 28000
 
-    # Case 2: Same as case (1) but the max_size_gib is specified.
```
homatthew: The max side gib test cases are no longer correct because they are handleded by min_count / node density in #171

service_capacity_modeling/models/common.py
```diff
@@ -717,6 +751,65 @@ def merge_plan(
...
+    @staticmethod
+    def for_components(buffer: Dict[str, Buffer], components: List[str]):
+        scale = 1.0
+        preserve = False
+        ceiling = None
+        floor = None
+
+        for bfr in buffer.values():
+            if any(component in components for component in bfr.components):
+                if bfr.intent == BufferIntent.scale:
```
ramsrivatsa, line 774:
```
        if bfr.intent in (BufferIntent.scale, BufferIntent.scale_up, BufferIntent.scale_down):
            scale = combine_buffer_ratios(scale, bfr.ratio)
```

service_capacity_modeling/models/common.py
```diff
@@ -717,6 +751,65 @@ def merge_plan(
...
+    def calculate_requirement(
+        self,
+        current_usage: float,
+        existing_capacity: float,
+        desired_buffer_ratio: float = 1.0,
+    ) -> float:
+        if self.preserve:
+            return existing_capacity
+
+        if self.scale is None:
```
ramsrivatsa, line 798:
```
requirement = (
    self.scale * current_usage * desired_buffer_ratio
    if self.scale is not None
    else current_usage * desired_buffer_ratio
)
```

Easier to understand.

service_capacity_modeling/models/common.py
```diff
@@ -806,115 +972,41 @@ def get_memory_from_current_capacity(
...
             reference_shape,
         )
-        needed_network_mbps = get_network_from_current_capacity(
-            current_capacity, buffers
-        )
+
+        needed_network_mbps = requirement.network
+        needed_disk_gib = requirement.disk_gib
+
+        # TODO: Memory should also be calculated via the requirement class
```
ramsrivatsa, line 1001: I think you can move the existing version of memory also into the `RequirementFromCurrentCapacity` class. The reason why I say that is, current memory would be precomputed and given directly. 

For cass, AG cass should do that and send it to you. This implementation would not change.

homatthew:
Rebase + Address coments

ramsrivatsa (approved):
LGTM. Please merge.

PR #170, opened 2025-08-19
Support reference_shape based on current cluster instance

homatthew:
This was a missing feature that causes the requirements for current clusters to provision less cores than expected because it would convert the number of cores based on the reference shape

Anytime we do a right-sizing (i.e. we know the current cluster instance), we should be normalizing requirements know that the usage is based on the current instance type and not the nameless default reference instance.

Here's what happens in practice:
```
default_reference_shape = Instance(
    name="default_reference_shape",
    cpu=4,
    # Much of our benchmarking was carried out in the 5th gen i2 shape, can
    # adjust this up once we go through and fix all the latency distributions
    cpu_ghz=2.3,
    cpu_ipc_scale=1.0,
    ram_gib=8.0,
    net_mbps=1000,
)

==> CPU utilization = 25%
```

Implicitly, we are using 1 core of this 2.3 ghz 1.0 ipc scale instance.

But what if the node that this is running on is actually larger? Then the actual load we are receiving 

Let's take an extreme example:
```
i4i.4xlarge
16 CPU
3.1 GHZ
1.0 IPC scale
```

25% of this CPU represents a significantly larger load so we will be under provisioning load when right-sizing.

### DIFF
--- a/service_capacity_modeling/interface.py
+++ b/service_capacity_modeling/interface.py
@@ -876,13 +876,22 @@ class CapacityDesires(ExcludeUnsetModel):
 
     @property
     def reference_shape(self) -> Instance:
-        if (
-            self.current_clusters
-            and self.current_clusters.zonal
-            and self.current_clusters.zonal[0].cluster_instance
-        ):
-            return self.current_clusters.zonal[0].cluster_instance
-        # TODO: Support regional reference shapes from the current cluster
+        if not self.current_clusters:
+            return default_reference_shape
+
+        zonal, regional = (self.current_clusters.zonal, self.current_clusters.regional)
+        if zonal and regional:
+            raise ValueError(
+                "The current cluster should not have both "
+                "zonal and regional instances. They're mutually exclusive."
+            )
+
+        if zonal and zonal[0].cluster_instance:
+            return zonal[0].cluster_instance
+
+        if regional and regional[0].cluster_instance:
+            return regional[0].cluster_instance
+
         return default_reference_shape
 
     def merge_with(self, defaults: "CapacityDesires") -> "CapacityDesires":
--- a/service_capacity_modeling/interface.py
+++ b/service_capacity_modeling/interface.py
@@ -876,7 +876,13 @@ class CapacityDesires(ExcludeUnsetModel):
 
     @property
     def reference_shape(self) -> Instance:
-        # TODO: this should use the shape from current clusters if it is there
+        if (
+            self.current_clusters
+            and self.current_clusters.zonal
+            and self.current_clusters.zonal[0].cluster_instance
+        ):
+            return self.current_clusters.zonal[0].cluster_instance
+        # TODO: Support regional reference shapes from the current cluster
         return default_reference_shape
 
     def merge_with(self, defaults: "CapacityDesires") -> "CapacityDesires":
--- a/tests/netflix/test_cassandra.py
+++ b/tests/netflix/test_cassandra.py
@@ -1,6 +1,7 @@
 import pytest
 
 from service_capacity_modeling.capacity_planner import planner
+from service_capacity_modeling.hardware import shapes
 from service_capacity_modeling.interface import (
     AccessConsistency,
 )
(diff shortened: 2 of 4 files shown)


tests/netflix/test_cassandra.py
```diff
@@ -430,7 +432,7 @@ def test_plan_certain(self):
             cluster_instance_name="i4i.8xlarge",
             cluster_instance_count=Interval(low=8, mid=8, high=8, confidence=1),
             cpu_utilization=Interval(
-                low=10.12, mid=13.2, high=14.194801291058118, confidence=1
+                low=10.12, mid=13.1, high=14.194801291058118, confidence=1
```
homatthew, line 435: Updated this because 13.2 would select i4i.8xlarge still as the top choice. 

It's interesting that 13.2 isn't enough to scale down. If this is a concern, I can look a little deeper to figure out the explanation for why we don't scale down even at 13.2 CPU usage

chengw-netflix: what's the default success buffers defined? is it 2x? does it mean as long as the CPU is below 20% it should be right sized to half?

homatthew: Default success buffer for "compute", which includes both memory and CPU is 1.5x and a 2x background CPU buffer. The math should be:

Let's take 1 example for just `i4i.8xlarge -> i4i.4xlarge`.

Assume we need a 25% headroom for an instance and 
```
let cluster = i4i.8xlarge (32 vcpu), 8 nodes per asg

(1 - ((1 - cpu_reserved_headroom) / 3)

buffer_adjusted_headroom_per_instance = (1 - cpu_headroom(i4i.4xlarge)) / 3
                  == (1 - 0.205) / 3 
                  == 0.795 / 3  ~= .26

cpu_headroom = 1 - buffer_adjusted_headroom
                 == 1 - 0.26
                 == 0.74

target_cpu  = 1 - headroom 
                == 0.26

effective_cpu_util = actual_cpu / target_cpu
                 == .131 / 0.26
                 == 0.5038

total_cluster_cpu = 32 vcpu * 8 nodes == 256
required_cpu == int(effective_cpu_util * compute_buffer *  total_cluster_cpu)
           == int(0.50308 * 3 * 256)
           == int(128.96) 
           == int(128) cores
```

In other words, the answer is *it depends* on what headroom because we need the effective CPU usage to be 33%. In this specific case, ~26% is the break-even point where we should scale up vs scale down. But because we tend to scale in orders of 2 (both vertically and horizontally), anything around `26 / 2 == 13`

jolynch (approved):
Nice, I think this makes sense that when provided a current cluster we should be taking any latency claims with that shape in mind.

service_capacity_modeling/interface.py
```diff
@@ -876,7 +876,13 @@ class CapacityDesires(ExcludeUnsetModel):
 
     @property
     def reference_shape(self) -> Instance:
-        # TODO: this should use the shape from current clusters if it is there
+        if (
+            self.current_clusters
+            and self.current_clusters.zonal
```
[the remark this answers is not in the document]
homatthew: The reason I left regional unset is because I wasn't sure the blast radius of also supporting regional. Is it things like stateless and aurora?

tests/netflix/test_cassandra.py
ramsrivatsa: @jolynch [This](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/main/service_capacity_modeling/models/common.py#L886-L890) is the only play we normalize cores. In other words, we normalize to the target instance type. 

Hence, when performing `compute_stateful_zone` we can directly compare needed_cores with that number of cores * target instance count (as done [here](https://github.com/Netflix-Skunkworks/service-capacity-modeling/blob/main/service_capacity_modeling/models/common.py#L385-L394)). Based on my read we are not normalizeing twice.

jolynch: Sounds great

PR #171, opened 2025-08-22
Bug fix: Handle node density with buffer via min count

homatthew:
**What changed:**
- Added `compute_max_data_per_node()` function to enforce per-node data limits
- Replaced hardcoded buffer ratios with configurable `default_buffers()` per model
- Models now respect `max_local_disk_gib` constraints when sizing clusters. In the past many clusters were over provisioned and allocating much less than this amount because buffer was never applied
- Added helper functions to centralize min node count logic

**Why:**
- Prevents overloading individual nodes with too much data
- Makes buffer ratios configurable instead of magic numbers
- Ensures cluster sizing considers both capacity needs and node density limits

### DIFF
--- a/service_capacity_modeling/models/org/netflix/elasticsearch.py
+++ b/service_capacity_modeling/models/org/netflix/elasticsearch.py
@@ -27,6 +27,7 @@ from service_capacity_modeling.interface import RegionContext
 from service_capacity_modeling.interface import Requirements
 from service_capacity_modeling.interface import ZoneClusterCapacity
 from service_capacity_modeling.models import CapacityModel
+from service_capacity_modeling.models.common import compute_density_gib
 from service_capacity_modeling.models.common import compute_stateful_zone
 from service_capacity_modeling.models.common import normalize_cores
 from service_capacity_modeling.models.common import simple_network_mbps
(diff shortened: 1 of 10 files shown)


service_capacity_modeling/models/org/netflix/evcache.py
```diff
@@ -297,16 +298,21 @@ def reserve_memory(instance_mem_gib):
         adjusted_disk_io_needed = 1.4 * adjusted_disk_io_needed
         read_write_ratio = reads_per_sec / (reads_per_sec + writes_per_sec)
 
+    needed_disk_gib = int(requirement.disk_gib.mid)
+    density_gib = compute_density_gib(
+        instance,
+        drive,
+        1,
+        max_local_disk_gib=max_local_disk_gib,
```
homatthew, line 306: I'm uncertain on if this is correct because EV cache also adjusts the count based on the reads / write IO. If someone is more familiar with this opinions are appreciated

service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -176,6 +181,23 @@ class NflxElasticsearchArguments(BaseModel):
 
 
 class NflxElasticsearchDataCapacityModel(CapacityModel):
+    @staticmethod
+    def default_buffers() -> Buffers:
+        return Buffers(
+            default=Buffer(ratio=1.5),
+            desired={
+                "storage": Buffer(ratio=1.33, components=[BufferComponent.storage]),
```
ramsrivatsa, line 189: I think the default buffers should be 1.33 desired is passed on.

PR #258, opened 2026-04-09
feat: add node-count bottleneck breakdown to compute_stateful_zone

homatthew:
## Summary
- Add opt-in `required_nodes_by_type` and `count_bottleneck` cluster params from `compute_stateful_zone`
- Keep explainability off by default for existing callers
- Split the attached-drive EBS IOPS overflow fix into its own atomic commit so the behavior change is reviewable on its own

## Commits
- `feat: add opt-in node-count bottleneck breakdown`
- `fix: recalculate EBS IOPS after overflow instead of scaling by ratio`

## Test plan
- [x] `tox -e py310 -- tests/test_resource_counts.py`
- [x] Added attached-drive IOPS overflow regression coverage

> Stack 1/3: base of #259, then #260.

### DIFF
--- a/service_capacity_modeling/models/common.py
+++ b/service_capacity_modeling/models/common.py
@@ -191,8 +191,12 @@ def _attached_drive_plan(
     if (read_io + write_io) > drive.max_io_per_s:
         ratio = (read_io + write_io) / drive.max_io_per_s
         count_disk_iops = math.ceil(effective_count * ratio)
-        read_io = utils.next_n(read_io * ratio, n=200)
-        write_io = utils.next_n(write_io * ratio, n=200)
+        iops_count = max(cluster_size(count_disk_iops), min_count)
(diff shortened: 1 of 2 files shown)


service_capacity_modeling/models/common.py
```diff
@@ -659,6 +654,15 @@ def compute_stateful_zone(  # pylint: disable=too-many-positional-arguments
         instance=instance,
         attached_drives=attached_drives,
         annual_cost=cost,
+        cluster_params={
```
homatthew: Arguably I could introduce a new field for context, I'll punt this to a future refactor since some other code uses cluster params for context.

homatthew:
Superseded by native upstream-branch PR #265. Closing this cross-repo version so the stack is reviewed through same-repo upstream branches.

PR #259, opened 2026-04-09
feat: surface Cassandra count bottleneck details and write-buffer caps

homatthew:
## Summary
- Surface `required_nodes_by_type` and `count_bottleneck` in Cassandra cluster-size excuses
- Use bottleneck wording in excuse text so callers can see why the computed count missed the required topology
- Cap `write_buffer_gib` from memory-only derived buffers when current cluster state is available
- Hydrate `cluster_instance` from name for the common name-only current-cluster path
- Keep explainability consumer and write-buffer behavior in separate atomic commits

## Commits
- `feat: expose Cassandra count bottleneck details`
- `feat: cap Cassandra write buffer from current memory policy`

## Test plan
- [x] `tox -e py310 -- tests/netflix/test_cassandra_resource_counts.py tests/netflix/test_cost_regression.py`
- [x] Memory scale-down cap path covered in `tests/netflix/test_cassandra_resource_counts.py`

> Stack 2/3: depends on #258. Merge order: #258 → #259 → #260.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/cassandra.py
+++ b/service_capacity_modeling/models/org/netflix/cassandra.py
@@ -46,6 +46,8 @@ from service_capacity_modeling.interface import ServiceCapacity
 from service_capacity_modeling.models import CapacityModel
 from service_capacity_modeling.models import CostAwareModel
 from service_capacity_modeling.models import RANK_PENALTIES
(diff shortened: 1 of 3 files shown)


homatthew:
Superseded by native upstream-branch PR #266. Closing this cross-repo version so the stack is reviewed through same-repo upstream branches.

PR #261, opened 2026-04-13
feat: expand uncertain planner explainability surface

homatthew:
## What am I trying to do?

Expand the uncertain planner explainability surface so callers can get a dedicated explained result object and richer regret breadcrumbs without reverse-engineering the raw planner output.

## Why did I do it this way?

I moved uncertain explainability onto `planner.plan_explained(...)` and made `planner.plan(...)` return only the `UncertainCapacityPlan`. This is an intentional API break: the old `explain` flag is removed. The new explained surface carries structured regret details, per-model regret summaries, and counted excuses so downstream consumers can narrate why the least-regret plans win and how often alternatives fail across sampled worlds.

I also tightened excuse grouping so distinct bottlenecks or tag sets are not silently merged into the same counted explanation bucket.

## Stack position

Base branch for the uncertain stack. This PR targets `main` directly. `#262` is intended to stack on top of this branch.

## Are there any tests?

Yes. Verified with:
- `tox -e py312 -- tests/test_explainability.py tests/netflix/test_cassandra_explainability.py tests/test_reproducible.py`

Coverage includes:
- `planner.plan_explained()` return type and wrapper behavior
- richer `PlanExplanation` fields (`regret_details_by_model`, `regret_summaries_by_model`, `excuse_counts_by_model`)
- counted-excuse grouping behavior
- reproducibility of uncertain planning output

```python
from service_capacity_modeling.capacity_planner import planner

explained = planner.plan_explained(
    model_name="org.netflix.cassandra",
    region="us-east-1",
    desires=desires,
    extra_model_arguments=extra,
)

explained.plan
explained.excuses
explained.excuse_summary
explained.least_regret_summaries
explained.plan.explanation.regret_details_by_model
explained.plan.explanation.regret_summaries_by_model
```

### DIFF
--- a/notebooks/kafka_demo.ipynb
+++ b/notebooks/kafka_demo.ipynb
@@ -88,28 +88,7 @@
    "id": "f07ab2d2",
    "metadata": {},
    "outputs": [],
-   "source": [
-    "from service_capacity_modeling.capacity_planner import planner\n",
-    "from service_capacity_modeling.models.org import netflix\n",
-    "\n",
-    "# Load up the Netflix capacity models\n",
-    "planner.register_group(netflix.models)\n",
-    "model_name = \"org.netflix.kafka\"\n",
-    "\n",
-    "\n",
-    "cap_plan = planner.plan(\n",
-    "    model_name=model_name,\n",
-    "    region=\"us-east-1\",\n",
-    "    desires=kafka_desires,\n",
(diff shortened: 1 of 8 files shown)


homatthew:
Superseded by #269, which uses an upstream branch head and a cleaner single-PR diff.

PR #262, opened 2026-04-15
test: capture uncertain planner regression baselines

homatthew:
## What am I trying to do?

Add regression coverage for the rebuilt uncertain planner surface, including both the compact uncertain plan snapshots and the richer explained uncertain summaries.

## Why did I do it this way?

This is the fixture and regression layer for the uncertain stack. The planner/API changes live in `#261`; this PR captures those outputs in stable baseline files so future changes show up as intentional diffs instead of silent drift.

I kept the branch focused on capture and regression artifacts:
- extend `capture_baseline_costs.py` to serialize uncertain and uncertain-explained outputs
- add uncertain regression tests
- add uncertain explained regression tests
- pin `tox -e capture-baseline` to Python 3.11 so local regeneration matches the canonical environment

## Stack position

Stacks on `#261` (`uncertain-explained`). This PR should be reviewed after the explainability API layer.

## Are there any tests?

Yes. Verified with:
- `tox -e capture-baseline`
- `tox -e py312 -- tests/netflix/test_uncertain_regression.py tests/netflix/test_uncertain_explained_regression.py`

Run `tox -e capture-baseline` to regenerate uncertain baseline fixtures after intentional planner changes. The regression suites will automatically compare current output to:
- `baseline_uncertain.json`
- `baseline_uncertain_explained.json`

### DIFF
--- a/service_capacity_modeling/tools/capture_baseline_costs.py
+++ b/service_capacity_modeling/tools/capture_baseline_costs.py
@@ -34,6 +34,9 @@ from service_capacity_modeling.interface import (
     QueryPattern,
 )
 
+BASELINE_UNCERTAIN_SIMULATIONS = 16
+BASELINE_UNCERTAIN_NUM_RESULTS = 3
+
 
 def _format_cluster(cluster: ClusterCapacity, deployment: str) -> dict[str, Any]:
     """Format a single cluster's details."""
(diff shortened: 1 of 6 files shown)


homatthew:
Superseded by #270, which stacks on the new upstream explainability PR and keeps only the plain uncertain baseline fixture.

PR #263, opened 2026-04-15
feat: enrich uncertain explained output for LLM deep dives

homatthew:
> Depends on #262

## What am I trying to do?
Extend the uncertain planner's explained surface so an LLM can explain more than the deterministic-looking winner. This adds regret summaries, per-model regret component breadcrumbs, and counted excuses so consumers can answer questions like why the least-regret choice wins and how often alternate shapes fail across sampled worlds.

## Why did I do it this way?
I kept the lower stochastic baseline capture in a separate prerequisite PR and layered the richer explainability on top. The new payload keeps the existing `excuses` and `family_graph` shape, but adds explicit summaries so consumers do not need to reverse-engineer regret or infer simulation frequency from deduped excuses.

## Are there any tests?
Yes. I added `tests/netflix/test_uncertain_explained_regression.py`, extended Cassandra explainability coverage, regenerated `baseline_uncertain_explained.json`, and verified with `tox -e capture-baseline` plus `tox -e py312 -- tests/netflix/test_cassandra_explainability.py tests/netflix/test_uncertain_regression.py tests/netflix/test_uncertain_explained_regression.py tests/test_reproducible.py`.

Call `planner.plan_explained(...)` as before, then read `least_regret_summaries`, `excuse_summary`, and the new `plan.explanation.regret_details_by_model` / `regret_summaries_by_model` fields to narrate deeper uncertain-planner tradeoffs.

### DIFF
--- a/.github/workflows/python-build.yml
+++ b/.github/workflows/python-build.yml
@@ -6,6 +6,18 @@ on:
     tags: ["v*"]
   pull_request:
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.11"
+    - name: Install Tox and any other packages
+      run: pip install tox
+    - name: Run linters via pre-commit
+      run: tox -e pre-commit
   build:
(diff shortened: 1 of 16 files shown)


homatthew:
Superseded by #261 and #262.

PR #264, opened 2026-04-16
docs: add repo agent workflow notes

homatthew:
## What am I trying to do?

Add repo-specific workflow notes so agents do not accidentally create misleading stacked PRs while touching planner code, baseline capture, or generated snapshots.

## Why did I do it this way?

This stays as a standalone docs/process PR so the workflow guidance can merge independently from the uncertain planner code. The most important addition is a clear warning about a common stacked-PR failure mode: a child PR can target the right parent branch by name while still being based on the wrong tip or wrong remote copy of that parent, which makes the PR graph look right while the commit DAG is wrong.

## Stack position

Standalone. This PR targets `main` directly and is not intended to carry planner behavior changes.

## Are there any tests?

No code-path tests. Verified with `tox -e pre-commit` during commit.

Read `AGENTS.md` before rewriting a stack, regenerating baselines, or splitting planner work across multiple PRs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,6 +11,9 @@
 - Keep stacked branches linear.
 - Each PR branch should contain only the commits for that layer relative to its base branch.
 - If a branch picks up extra history, rebuild it from the intended base instead of dragging parent commits forward.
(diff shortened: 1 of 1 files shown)


homatthew:
Superseded by #268, which carries the same docs guidance on an upstream branch head.

PR #266, opened 2026-04-16
Fix Cassandra memory sizing and resource counts

homatthew:
## What am I trying to do?

Fix Cassandra capacity recommendations so memory sizing no longer scales page-cache demand linearly with total state size. This makes 4xlarge-class instances viable for write-heavy workloads when heap, reserves, and useful page cache fit in node RAM.

### Bounded Cassandra Memory

Cassandra RAM is partitioned into heap, physical reserves, and effective page cache. Page-cache demand is still driven by working-set bounds, but it is capped by what the planned cluster can usefully serve instead of treating large state size as a requirement to cache a flat percentage of disk.

### Single Memory Path

The old legacy/experimental split is removed. The unified path applies SLO, RPS, and observational bounds consistently, then emits explicit page-cache demand, linear demand, effective budget, and write-buffer demand.

### Reviewability

Memory helper containers are frozen Pydantic models, generated Cassandra baselines were refreshed, and regression coverage checks the m7a.4xlarge/r7a.4xlarge fit behavior.

## Are there any tests?

Yes. Focused Cassandra planner tests cover the memory model, resource counts, and 4xlarge fit regression. `tox -e pre-commit` also passed before push.

No caller API change is required. Cassandra capacity planning uses the bounded memory model by default when producing recommendations.

## Follow-ups

Unknown RPS/WPS policy and preserve-memory rework are separate follow-ups. PR 267 is stacked on this branch and handles derived buffer normalization.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/cassandra.py
+++ b/service_capacity_modeling/models/org/netflix/cassandra.py
@@ -1,6 +1,7 @@
 # pylint: disable=too-many-lines
 import logging
 import math
+from functools import lru_cache
 from typing import Any
 from typing import Callable
 from typing import Dict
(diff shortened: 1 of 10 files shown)


service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -288,8 +291,9 @@ def _estimate_cassandra_requirement(
     max_rps_to_disk: int,
     zones_per_region: int = 3,
     copies_per_region: int = 3,
+    required_cluster_size: Optional[int] = None,
     experimental_memory_model: bool = False,
-    max_page_cache_gib: float = 32.0,
+    max_page_cache_gib: float = 28.0,
```
homatthew: This max page cache gib has a poor interaction with the other thing constriants. 32 + 30 + 3 > 64 which means most use cases cant use a m6id.2xlarge (64 gb ram).

PR #270, opened 2026-04-17
test: capture uncertain planner regression baselines

homatthew:
## What am I trying to do?
Capture regression baselines for uncertain planner output while avoiding the oversized explained snapshot fixture.

## Why did I do it this way?
This branch is the fixture layer on top of the explainability changes. It keeps only `baseline_uncertain.json` and the plain uncertain regression test. The explained snapshot path was removed because it creates too much churn for small changes.

## Are there any tests?
Yes.
- `tox -e capture-baseline`
- `tox -e py312 -- tests/netflix/test_uncertain_regression.py`

Run `tox -e capture-baseline` after intentional uncertain planner changes. The checked-in fixture remains `baseline_uncertain.json` only.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -62,6 +62,9 @@ from service_capacity_modeling.stats import interval_percentile
 
 logger = logging.getLogger(__name__)
 
+IntervalSampler = Callable[[Interval, str, int], Sequence[Interval]]
+IntervalPercentileSampler = Callable[[Interval, Sequence[int]], Sequence[Interval]]
+
 
 class PlannerArguments(ExcludeUnsetModel):
     """Planner-level configuration, separate from model-specific extra_model_arguments.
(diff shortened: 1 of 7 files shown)


homatthew:
NOTE: As of today, scipy 1.17 was released on Feb 23rd, and there's smalling floating point differences in the output between linux / macos. I'll fix forward knowing that it's beneficial to have some observability on the baseline cost for uncertain plans. 

Codex:

> 1. We fit beta/gamma distributions from intervals using SciPy numeric routines.
>   2. That fitting can produce slightly different distribution parameters across SciPy versions, BLAS/libm/compiler/platform details, or tolerance behavior.
>   3. The random seed then samples deterministically from whatever fitted distribution it was given.
>   4. If the fitted distribution differs, the same seed can produce different sampled values.
> 
>   So the seed is deterministic, but it does not freeze the fitted distribution.
> 
>   For this CI failure, the obvious culprit was version resolution:
> 
>   - CI Python 3.12 installed scipy==1.17.1.
>   - Python 3.10 CI had scipy==1.15.3.
>   - With scipy<1.17, the churn disappeared locally and CI passed.
> 
>   So I would phrase it carefully:
> 
>   “Same seed gives deterministic samples for a fixed SciPy/Numpy/platform stack. The planner’s interval-to-distribution fitting is not guaranteed byte-stable across SciPy versions,
>   and may also drift across platforms. We saw a concrete SciPy-version difference here.”

PR #272, opened 2026-04-21
Add P-family GPU instance shapes and fix per-xlarge IOPS scaling

homatthew:
## Summary

Fix shape-generation math that drifts on flagship GPU families, then add P-family AWS shapes: `p4d`, `p4de`, `p5`, `p5en`, `p6-b200`.

### Tool changes

- **IOPS scales by storage GiB, not xlarge count.** Renames `xl_iops` → `iops_per_gib`, `--xl-iops` → `--iops-per-gib`, `IOPerformance.{xl_read_iops, xl_write_iops}` → `{read_iops_per_gib, write_iops_per_gib}`. Old `rate × xlarge_count` broke on families like `p5` where `p5.4xlarge` has 1×3.8 TB drive and `p5.48xlarge` has 8× — storage grows 8x, xlarge count grows 12x. The per-GiB rate is an approximation (~0% for uniform-drive families, <1% for linearly-scaling, up to ~3% for non-linear like c5d 12xl/24xl — all inside capacity-planning noise; see inline comment in `aws_iops_per_gib`).
- **Network sums every EFA card.** Was reading `NetworkCards[0]` only; P-family ships with 4–32 cards, so bandwidth was understated up to 32×.
- **`_parse_family` accepts multi-numeric names** like `p6-b200`.
- **`--lifecycle` flag plumbed through** `auto_shape.py` and `generate_missing.py`. RDS `db.*` families warn and ignore it.
- **Non-P committed JSONs stay unchanged** — single network card, linear scaling, same numbers under new math.

### P-family shapes

| instance | vCPU | RAM | net | drives | lifecycle |
|---|---|---|---|---|---|
| `p4d.24xlarge`, `p4de.24xlarge` | 96 | 1099 GiB | 400 Gbps | 8×1 TB | alpha |
| `p5.4xlarge` | 16 | 244 GiB | 100 Gbps | 1×3.8 TB | alpha |
| `p5.48xlarge`, `p5en.48xlarge` | 192 | 1953 GiB | 3200 Gbps | 8×3.8 TB | alpha |
| `p6-b200.48xlarge` | 192 | 1953 GiB | 1600 Gbps | 8×3.8 TB | alpha |

All `alpha`. `cpu_ipc_scale=None` → SMT heuristic → 1.0 uniformly. The `alpha` lifecycle keeps these shapes out of default CPU-bound workload plans (Cassandra, EVCache, Kafka); GPU-aware workloads can opt in by name or by `lifecycles=(Lifecycle.alpha, ...)`. Promote to `beta`/`stable` once benchmarked.

### Planner impact

`CapacityPlanner` defaults to `(Lifecycle.stable, Lifecycle.beta)` ([capacity_planner.py:475](service_capacity_modeling/capacity_planner.py#L475)). Alpha shapes filtered out of default Cassandra/EVCache/Kafka plans. Opt in via `lifecycles=(Lifecycle.alpha, ...)` or `allowed_names=[...]`.

### Tests & cleanup

- `test_memory_proportional_to_cpu` split into two parametrized tests: strict uniform-ratio for most families, non-increasing-as-size-grows for `ASYMMETRIC_RAM_FAMILIES = {"p5"}` (AWS pads p5.4xl to 16 GiB/vCPU vs p5.48xl's 10.67).
- `guess_iops` → `guess_iops_per_gib` rename.
- `i7i` IOPS consolidated to `aws_iops_per_gib` defaults table.

## Follow-ups

- [ ] Pricing for `p5en.48xlarge` and `p6-b200.48xlarge` missing from `3yr-reserved_ec2.json`; `persistence_prod_cde` role lacks `pricing:GetProducts`.
- [ ] Catalog refresh PR (`--force` regen) to apply per-GiB math across all families.
- [ ] Pre-existing `convert_mib_to_gib` base-10 bug (~4.6% RAM under-count).

## Test plan

- [x] `pytest tests/` — 630 passed, 1 skipped
- [x] P IOPS match AWS docs within rounding (reads/writes): `p4d` 1,999,848/1,599,730; `p5.4xl` 549,961/274,980; `p5.48xl`/`p6-b200.48xl` 4,399,685/2,199,842
- [x] Network aggregation cross-checked against `describe_instance_types` per shape
- [x] `i7i`: `deduce_io_perf("i7i", ...)` returns identical tuple via defaults vs explicit entry

### DIFF
--- a/service_capacity_modeling/hardware/profiles/shapes/aws/auto_c5d.json
+++ b/service_capacity_modeling/hardware/profiles/shapes/aws/auto_c5d.json
@@ -11,8 +11,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 47,
-        "read_io_per_s": 20000,
-        "write_io_per_s": 9000,
+        "read_io_per_s": 20215,
+        "write_io_per_s": 9094,
         "single_tenant": false,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -35,8 +35,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 93,
-        "read_io_per_s": 40000,
-        "write_io_per_s": 18000,
+        "read_io_per_s": 39999,
+        "write_io_per_s": 17996,
         "single_tenant": false,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -59,8 +59,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 186,
-        "read_io_per_s": 80000,
-        "write_io_per_s": 36000,
+        "read_io_per_s": 79999,
+        "write_io_per_s": 35991,
         "single_tenant": false,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -83,8 +83,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 373,
-        "read_io_per_s": 160000,
-        "write_io_per_s": 72000,
+        "read_io_per_s": 160427,
+        "write_io_per_s": 72176,
         "single_tenant": false,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -107,8 +107,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 838,
-        "read_io_per_s": 360000,
-        "write_io_per_s": 162000,
+        "read_io_per_s": 360424,
+        "write_io_per_s": 162153,
         "single_tenant": true,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -131,8 +131,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 1676,
-        "read_io_per_s": 480000,
-        "write_io_per_s": 216000,
+        "read_io_per_s": 720848,
+        "write_io_per_s": 324306,
         "single_tenant": true,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -155,8 +155,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 1676,
-        "read_io_per_s": 720000,
-        "write_io_per_s": 324000,
+        "read_io_per_s": 720848,
+        "write_io_per_s": 324306,
         "single_tenant": true,
         "read_io_latency_ms": {
           "low": 0.08,
@@ -179,8 +179,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 3353,
-        "read_io_per_s": 960000,
-        "write_io_per_s": 432000,
+        "read_io_per_s": 1442125,
+        "write_io_per_s": 648806,
         "single_tenant": true,
         "read_io_latency_ms": {
           "low": 0.08,
--- a/service_capacity_modeling/hardware/profiles/shapes/aws/auto_p4d.json
+++ b/service_capacity_modeling/hardware/profiles/shapes/aws/auto_p4d.json
@@ -1,0 +1,28 @@
+{
+  "instances": {
+    "p4d.24xlarge": {
+      "name": "p4d.24xlarge",
+      "cpu": 96,
+      "cpu_cores": 48,
+      "cpu_ghz": 3.0,
+      "cpu_ipc_scale": 1.0,
+      "ram_gib": 1098.63,
+      "net_mbps": 100000.0,
+      "drive": {
+        "name": "ephem",
+        "size_gib": 7451,
+        "read_io_per_s": 1999848,
+        "write_io_per_s": 1599730,
+        "single_tenant": true,
+        "read_io_latency_ms": {
+          "low": 0.08,
+          "mid": 0.125,
+          "high": 0.2,
+          "confidence": 0.9,
+          "minimum_value": 0.07,
+          "maximum_value": 2.0
+        }
+      }
+    }
+  }
+}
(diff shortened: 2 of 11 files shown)


jasonk000:
Assume you cross checked the change in volume data that nothing else changed? Might be worth a generate --force to see whether any other instance types changed.

homatthew:
@jasonk000 Running --force causes some changes to the final shape within a certain margin. I could some different strategies to make it more accurate, but it's within the margin of general capacity planning error so I am willing to accept it. 

See the below Claude summary of things

---------------

### Category 1: EFA network aggregation fix (expected, this PR)

Only affects multi-EFA-card instances — P-family. Other families have a single NetworkCard so they see no change.

### Category 2: IOPS/GiB vs per-xlarge math — bounded drift

Every ephemeral-storage family shifts slightly because the per-GiB rate is an approximation of AWS's actual per-drive-SKU rate. Per-family max drift vs AWS published docs after a hypothetical `--force` regen:

| family category | examples | max drift |
|---|---|---|
| uniform drives | p4d, p5, p5en, p6-b200, trn2u, dl1, f2 | 0% (exact — single drive SKU per family) |
| linearly-scaling | m6id, r6id, c6id, m6idn, r6idn, i4i, i7i | <1% (within rounding, ≤±8 IOPS) |
| non-linear | c5d 12xl/24xl | ~3% |
| pre-existing AWS irregularities | m5d 600 GB, i3 writes | ~3% (same drift committed JSONs already have under old math) |

**c5d is the interesting one.** Old `xl_iops × xlarge_count` produced:
- committed `c5d.12xlarge`: 480,000 read / 216,000 write
- AWS published: 700,000 / 340,000
- drift: **-31% / -36%**

That's because c5d.12xl uses 2×900 GB drives — the layout breaks with xlarge count. Under per-GiB math, c5d.12xl regenerates to 720,848 / 324,306 read/write — +3% / -5% vs AWS. Net improvement across every affected size.

### Why not fix in this PR

The --force regen would shift 15 files (most by <1%, some by 3%) as a side effect of the P-family work. I'd rather that diff land on its own so reviewers see what's changing and why, separate from the P-family addition. Filed as a follow-up in the PR description.

I also added an inline comment above `aws_iops_per_gib` in auto_shape.py (commit `60e9208`) spelling out the expected drift per family category and pointing at per-drive-SKU overrides as the escape hatch if any specific shape needs tighter accuracy. 0.5-3% sits inside capacity-planning noise, but the comment makes that choice visible.

Short version: cross-checked, drift is bounded and documented, c5d gets significantly more accurate under the new math, other families shift imperceptibly.

service_capacity_modeling/hardware/profiles/shapes/aws/auto_c5d.json
```diff
@@ -131,8 +131,8 @@
       "drive": {
         "name": "ephem",
         "size_gib": 1676,
-        "read_io_per_s": 480000,
-        "write_io_per_s": 216000,
+        "read_io_per_s": 720848,
+        "write_io_per_s": 324306,
```
homatthew, line 135: Breakdown of how each c5d size moved under the per-GiB math (vs. the previous committed values under old `xl_iops × xlarge_count`):

| size | drives | before | after | AWS docs | delta vs before |
|---|---|---|---|---|---|
| c5d.large | 1×50 GB | 20,000 / 9,000 | 20,215 / 9,094 | 20,000 / 9,000 | +1% / +1% |
| c5d.xlarge | 1×100 GB | 40,000 / 18,000 | 39,999 / 17,996 | 40,000 / 18,000 | ~0 |
| c5d.2xlarge | 1×200 GB | 80,000 / 36,000 | 79,999 / 35,991 | 80,000 / 37,000 | ~0 / ~0 |
| c5d.4xlarge | 1×400 GB | 160,000 / 72,000 | 160,427 / 72,176 | 175,000 / 75,000 | ~0 / ~0 |
| c5d.9xlarge | 1×900 GB | 360,000 / 162,000 | 360,424 / 162,153 | 350,000 / 170,000 | ~0 / ~0 |
| **c5d.12xlarge** | **2×900 GB** | **480,000 / 216,000** | **720,848 / 324,306** | **700,000 / 340,000** | **+50% / +50%** |
| **c5d.18xlarge** | **2×900 GB** | 720,000 / 324,000 | 720,848 / 324,306 | 700,000 / 340,000 | ~0 (math now matches 12xl since drive layout is identical) |
| **c5d.24xlarge** | **4×900 GB** | **960,000 / 432,000** | **1,442,125 / 648,806** | **1,400,000 / 680,000** | **+50% / +50%** |

The big shifts are at **12xl, 24xl** — c5d at these sizes uses extra drives rather than a bigger single drive, so `xl_iops × xlarge_count` under-counted by ignoring the drive-count multiplier. Per-GiB × total_gib captures it correctly.

18xl is unchanged because old math over-counted it (no corresponding drive bump beyond 12xl) to roughly the right number by accident; new math lands almost identical.

Small sizes (≤9xl) shift by <1 IOPS — that's just float-math rounding drift between the two formulations; same numbers.

All remaining drift vs AWS docs is ≤5%, inside capacity-planning noise, and documented in the `aws_iops_per_gib` comment in auto_shape.py.

homatthew:
>The --force regen would shift 15 files (most by <1%, some by 3%) as a side effect of the P-family work. I'd rather that diff land on its own so reviewers see what's changing and why, separate from the P-family addition. Filed as a follow-up in the PR description.

I changed my mind. I regenerated c5d :)

PR #273, opened 2026-04-25
[ODS-3502] Consider Existing Writer's CPU Utilization in Aurora Capacity Recommendation

agali-09:
Current Aurora capacity model used query-pattern signals and this might not be enough when a writer is already running. Relying only on those signals alone can still recommend an instance that is too small for the CPU the cluster is already using.

This PR makes the model use `CapacityDesires.current_clusters` when it is present. The existing writer’s CPU utilization is also included in the regional `current_clusters` input and hence recommendations respect observed writer CPU along with the query-pattern derived load.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/aurora.py
+++ b/service_capacity_modeling/models/org/netflix/aurora.py
@@ -11,6 +11,9 @@ from pydantic import Field
 
 from service_capacity_modeling.interface import AccessConsistency
 from service_capacity_modeling.interface import AccessPattern
+from service_capacity_modeling.interface import Buffer
+from service_capacity_modeling.interface import BufferComponent
+from service_capacity_modeling.interface import Buffers
 from service_capacity_modeling.interface import CapacityDesires
 from service_capacity_modeling.interface import CapacityPlan
 from service_capacity_modeling.interface import CapacityRequirement
@@ -30,11 +33,15 @@ from service_capacity_modeling.interface import RegionClusterCapacity
 from service_capacity_modeling.interface import RegionContext
 from service_capacity_modeling.interface import Requirements
 from service_capacity_modeling.models import CapacityModel
+from service_capacity_modeling.models.common import buffer_for_components
 from service_capacity_modeling.models.common import normalize_cores
 from service_capacity_modeling.models.common import simple_network_mbps
 from service_capacity_modeling.models.common import sqrt_staffed_cores
 
 logger = logging.getLogger(__name__)
+
+# Buffer component for P95 to provisioned headroom on existing writer CPU utilization.
+AURORA_WRITER_CPU_BUFFER = "aurora-writer-cpu"
 
 
 def _existing_writer_cpu_floor_cores(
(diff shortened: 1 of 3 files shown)


service_capacity_modeling/models/org/netflix/aurora.py
```diff
@@ -37,6 +37,29 @@
...
+    best = 0
+    for row in cc.regional:
+        ref = row.cluster_instance
+        cpu_pct = row.cpu_utilization.mid
+        if ref is None or ref.cpu <= 0 or cpu_pct < 1.0:
+            continue
+        # ref.cpu * (cpu_pct/100): Number of cores on the reference writer.
+        # An extra headroom of 1.15 is added above point utilization since
+        # cpu_pct is P95.
+        load_cores_on_ref = float(ref.cpu) * (cpu_pct / 100.0) * 1.15
```
homatthew: You may want the headroom to be configurable. You can leverage the compute resource + `buffer` fields to achieve this idiomatic within the code-base

homatthew:
CI failed

```
[INFO] This may take a few minutes...
ruff.....................................................................Passed
ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

1 file reformatted, 103 files left unchanged

pre-commit: exit 1 (18.03 seconds) /home/runner/work/service-capacity-modeling/service-capacity-modeling> pre-commit run --all-files pid=2425
  pre-commit: FAIL code 1 (34.89=setup[16.85]+cmd[18.03] seconds)
  evaluation failed :( (35.25 seconds)
```

Install the pre-commit hook and this shouldn't have any issues in the future

service_capacity_modeling/models/org/netflix/aurora.py
homatthew: One way to make this configurable without adding a new extra model argument is to wire up a default cpu buffer in the code, and then read the effective ratio from `desires.buffers`.

Cassandra does it this way: it defines default buffers in `NflxCassandraCapacityModel.default_buffers()` and then consumes the effective value with `buffer_for_components(...)` in `_get_cores_from_desires`:

```python
cpu_buffer = buffer_for_components(
    buffers=desires.buffers,
    components=[BACKGROUND_BUFFER],
)
needed_cores = math.ceil(sqrt_staffed_cores(desires) * cpu_buffer.ratio)
```

Aurora could do the same for this writer CPU headroom. Rough sketch:

```python
AURORA_WRITER_CPU_BUFFER = "aurora-writer-cpu"

@staticmethod
def default_buffers(writer_cpu_ratio: float = 1.15) -> Buffers:
    return Buffers(
        desired={
            AURORA_WRITER_CPU_BUFFER: Buffer(
                ratio=writer_cpu_ratio,
                components=[BufferComponent.cpu, AURORA_WRITER_CPU_BUFFER],
                explanation="Headroom over observed Aurora writer CPU utilization.",
            )
        }
    )
```

Then include that in `default_desires`:

```python
return CapacityDesires(
    ...,
    buffers=NflxAuroraCapacityModel.default_buffers(),
)
```

And pull it where the current 1.15 multiplier is used:

```python
writer_cpu_buffer = buffer_for_components(
    buffers=desires.buffers,
    components=[AURORA_WRITER_CPU_BUFFER],
)
load_cores_on_ref = ref.cpu * (cpu_pct / 100.0) * writer_cpu_buffer.ratio
```

That keeps the default behavior at 15%, but lets callers override the ratio via the existing `CapacityDesires.buffers.desired` merge path instead of baking the number permanently into the Aurora model.

agali-09: Thank you, done, I've updated the PR

PR #281, opened 2026-05-20
Fix mutual exclusivity of IPC gain and HT in deduce_cpu_ipc_scale

hsane-dev:
## Title
Fix auto_shape to multiply arch-ipc and no-HT factor in deduce_cpu_ipc_scale

## Problem
`deduce_cpu_ipc_scale()` in `auto_shape.py` treated architectural IPC and the
  no-HT factor as mutually exclusive due to an early return:

  When generate_missing.py passes --cpu-ipc-scale (e.g. GENOA_IPC=1.2995
  for m7a), the first branch always wins and the no-HT factor is never applied.

  _normalize_cpu() in common.py always divides by vCPU count, so
  cpu_ipc_scale must carry both factors — architectural IPC improvement AND
  the no-HT factor (1 vCPU = 1 full physical core)

## Solution
 Separate the assignment from the return so both factors always apply and return the product
  `return arch_ipc * ht_factor`
 
For HT-on instance families, there should be no change as a factor of 1.0 for HT shall be applied against the Skylake baseline.

## Testing
- Added 4 unit tests for deduce_cpu_ipc_scale covering all combinations of arch IPC present/absent × HT on/off

### DIFF
--- a/service_capacity_modeling/tools/auto_shape.py
+++ b/service_capacity_modeling/tools/auto_shape.py
@@ -143,13 +143,17 @@ def deduce_cpu_ipc_scale(
 ) -> float:
     """
     Deduce CPU IPC scale factor from vCPU and core counts.
-    If all vCPUs are full cores (no SMT), use 1.5, otherwise 1.0.
+    Combine IPC gain and HT as independent factors:
+        - arch_ipc: per-physical core IPC compared to baseline=Skylake (IPC = 1.0)
+        - ht_factor: hyperthreading impact, 1.5 when no SMT (vcpu = cpu_cores), 1.0 when SMT (vcpu = 2x cpu_cores)
+    Both factors are always applied
     """
+    arch_ipc = 1.0
     if cpu_perf is not None and cpu_perf.ipc_scale_factor is not None:
-        return cpu_perf.ipc_scale_factor
-    if vcpu_count == cpu_cores:
-        return 1.5
-    return 1.0
+        arch_ipc = cpu_perf.ipc_scale_factor
+    ht_factor = 1.5 if vcpu_count == cpu_cores else 1.0
+
+    return arch_ipc * ht_factor
 
 
 def convert_mib_to_gib(size_mib: float) -> float:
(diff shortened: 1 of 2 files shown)


argha-c (approved):
LGTM. The separation makes sense IMO.
One minor clarification on the math applied.

/cc @jolynch to cross-check

service_capacity_modeling/tools/auto_shape.py
```diff
@@ -143,13 +143,17 @@ def deduce_cpu_ipc_scale(
...
+    arch_ipc = 1.0
     if cpu_perf is not None and cpu_perf.ipc_scale_factor is not None:
-        return cpu_perf.ipc_scale_factor
-    if vcpu_count == cpu_cores:
-        return 1.5
-    return 1.0
+        arch_ipc = cpu_perf.ipc_scale_factor
+    ht_factor = 1.5 if vcpu_count == cpu_cores else 1.0
+
+    return arch_ipc * ht_factor
```
argha-c, line 156: Does this math(compounded benefit of h/t *and* ipc factor) hold true linearly in all benchmarks?

hsane-dev: This is generalized across applications and the math should generally hold true. Now ofcourse there would be exceptions but since we dont have a mechanism to read application wise perfspect metrics (now in Atlas) into this today, this is the best static math we can do.

jolynch (approved):
lg2m 🚢

jolynch:
Thank you @hsane-dev ! Do we need to regenerate any shapes or do we think this will have the intended effect naturally?

service_capacity_modeling/tools/auto_shape.py
```diff
@@ -143,13 +143,17 @@ def deduce_cpu_ipc_scale(
 ) -> float:
     """
     Deduce CPU IPC scale factor from vCPU and core counts.
-    If all vCPUs are full cores (no SMT), use 1.5, otherwise 1.0.
+    Combine IPC gain and HT as independent factors:
+        - arch_ipc: per-physical core IPC compared to baseline=Skylake (IPC = 1.0)
+        - ht_factor: hyperthreading impact, 1.5 when no SMT (vcpu = cpu_cores), 1.0 when SMT (vcpu = 2x cpu_cores)
+    Both factors are always applied
```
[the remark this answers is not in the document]
hsane-dev: Not required

hsane-dev:
@jolynch  We would likely have to regenerate the 8th gen shapes, but the 7th gen should be unaffected.

PR #282, opened 2026-06-04
Add i7ie instance family using auto_shape

jolynch:
Adds i7ie to unblock new launches.

```
.tox/dev/bin/python -m service_capacity_modeling.tools.auto_shape i7ie --output-path service_capacity_modeling/hardware/profiles/shapes/aws/
```

I imagine some tests will break, uploading now then will see what breaks to fix it.


homatthew:
Checked with eyeballs that the IOPs match
- https://docs.aws.amazon.com/ec2/latest/instancetypes/so.html#so_instance-store
image

service_capacity_modeling/tools/instance_families.py
```diff
@@ -114,6 +114,13 @@
         "cpu_turbo_single_ghz": 4.0,
         "cpu_turbo_all_ghz": 3.2,
     },
+    "i7ie": {
+        "iops_per_gib": None,  # falls back to aws_iops_per_gib["i7ie"] defaults
+        "io_latency_curve": "7th-gen-ephemeral",
+        "cpu_ipc_scale": EMERALD_RAPIDS_IPC,
+        "cpu_turbo_single_ghz": 4.0,
+        "cpu_turbo_all_ghz": 3.2,
```
homatthew, line 122: It's worth noting for future readers that these values are purely decorative and are not currently used anywhere. They are more or less hand written and come from the AWS page. 

The actual value used for characterization in the `auto_shape` tool are pulled directly from the aws EC2 api (2.4 ghz for i7i and 3.6 for i7ie).

homatthew:
@jolynch The reason that we are missing i7i and i7ie is that the pricing info is missing.

I'll tackle that in a separate PR

jolynch:
Yeah the auto-pricing fetch is broken for the newer "only pricing plan" shapes >_<

Let's see if I can patch that up to just take the pricing plan API and apply the 3-yr discount to the on-demand price or something or just estimate.

jolynch:
# Cost-regression baseline drift analysis

After adding `i7ie` and the Savings Plan pricing fallback, three
`tests/netflix/test_cost_regression.py::TestBaselineDrift` scenarios failed the
±1% total-cost tolerance. This documents what failed and why it is **expected
drift, not a regression**.

## Failures

```
FAILED kafka_500mib_throughput      baseline=$98,036.73  actual=$82,750.08  -15.59%
FAILED evcache_small_no_replication baseline=$31,265.13  actual=$28,550.73   -8.68%
FAILED kv_with_cache                baseline=$618,040.95 actual=$608,687.19  -1.51%
```

All three moved **down** (cheaper).

## Root cause

Each scenario now selects a newer instance family that only became selectable
through this change — either the brand-new `i7ie` family, or a Granite Rapids
family (`m8i`/`c8i`) that had **no Reserved Instance price** before and only got
one via the EC2 Instance Savings Plan gap-fill.

| Scenario | Now selects | Why newly available |
|---|---|---|
| `kafka_500mib_throughput` | `i7ie.large` ×32 ×3 zones | family added in this PR |
| `evcache_small_no_replication` | `m8i.xlarge` ×13 ×3 zones | SP gap-fill gave it a first price |
| `kv_with_cache` | `c8i.4xlarge` ×39 (java-app regional) | SP gap-fill gave it a first price |

For `kv_with_cache` only the `nflx-java-app.regional-clusters` bucket changed
($108,433 → $99,079); every other bucket (cassandra, evcache, network, backup)
is byte-for-byte identical, which is why its total only moved −1.5%.

## Why the new prices are trustworthy (not a mispricing pulling the planner)

The drift is only legitimate if the new prices are correct. Spot-check vs
instances.vantage.sh (us-east-1, Linux, 3yr All Upfront):

| Instance | Our $/hr | Vantage $/hr | Note |
|---|---|---|---|
| `i7ie.xlarge` | 0.1969 | 0.1969 | exact match |
| `c8i.4xlarge` | 0.2873 | 0.28727 | exact match |
| `i7ie.large` | 0.0984 | 0.1153 | below — we use deeper EC2 Instance SP |
| `m8i.xlarge` | 0.0836 | 0.09193 | below — we use deeper EC2 Instance SP |

Internal consistency confirms these are real EC2 Instance SP rates:
`i7ie.large` ($861.98/yr) is **exactly half** of `i7ie.xlarge` ($1,724.84/yr) —
the per-vCPU linear scaling EC2 Instance SP guarantees within a family (2 vCPU vs
4 vCPU). All discounts land in the expected 60–62% off on-demand range; none are
near-zero or otherwise anomalous.

## Conclusion / fix

Not a regression — the planner is correctly choosing genuinely-cheaper,
correctly-priced instances that did not exist (or had no price) when the baseline
was frozen. The frozen `baseline_costs.json` is simply stale and must be
recaptured so I did that via

```bash
tox -e capture-baseline
```

jolynch:
Ok, so this is interesting, adding in pricing SP gap-filled has now caused tests relating to the i7i to fail. I'm pretty sure because the ec2 instance API is returning 2.4GHz for the all core turbo on i7i, which is ... not accurate.

```
aws --region us-east-1 ec2 describe-instance-types --filters 'Name=instance-type,Values=i7i.xlarge*' | grep Ghz 
                "SustainedClockSpeedInGhz": 2.4,
```

i7ie reports 3.6, and both [vantage](https://instances.vantage.sh/aws/ec2/i7i.xlarge?currency=USD) and aws docs say i7i and i7ie are 3.2.

I think I should override both of them to 3.2 manually but maybe @hsane-dev or @jasonk000 can weigh in what the perf testing actually observed. I could just make auto_shape draw from cpu_turbo_all_ghz in the instance_families.py script (which afaik isn't used by auto_shape.py)

service_capacity_modeling/tools/instance_families.py
jolynch: As of the latest patch we'll use the turbo_all if "nflx" cpu flag is passed

homatthew: As a minor nit: I wish there was a more intuitive way to know that these fields are nflx provided and not from some API. 

But it's also not that hard to remember that this `.py` file is human generated. And then make the mental connection that the `nflx` flag would pull from here.

service_capacity_modeling/tools/auto_shape.py
```diff
@@ -308,11 +311,35 @@ def debug_log(msg: str) -> None:
...
         cpu_ipc_scale_factor = deduce_cpu_ipc_scale(vcpu_count, cpu_cores, cpu_perf)
 
+        # AWS's SustainedClockSpeedInGhz is inconsistent across families (base for
+        # some, turbo for others), which makes cpu_ghz * cpu_ipc_scale misleading
+        # for families that report a low sustained clock (e.g. i7i). With
+        # --cpu-freq-from nflx, use the curated cpu_turbo_all_ghz instead; with
+        # aws (the default) use AWS's reported clock.
+        aws_ghz = data["ProcessorInfo"]["SustainedClockSpeedInGhz"]
+        cpu_ghz = aws_ghz
+        if cpu_freq_from == "nflx":
```
homatthew, line 321: Could we make this a family level property somehow? I'd hate to remember which ones should be "nflx" sourced vs "aws" sourced if we had to regenerate the json files (e.g. adding a new characteristic).

service_capacity_modeling/tools/fetch_pricing.py
```diff
@@ -80,6 +80,75 @@ def fetch_ec2_pricing(pricing_client: Any) -> Dict[str, Dict[str, Union[float, s
...
+        for result in response.get("searchResults", []):
+            offering = result.get("savingsPlanOffering", {})
+            if offering.get("durationSeconds") != THREE_YEAR_SECONDS:
+                continue
+
+            instance_type = _savings_plan_instance_type(result)
+            if instance_type is None:
+                continue
+            # De-dupe defensively: keep the first 3yr match per instance type.
+            if instance_type in instances:
```
homatthew, line 139: I wouldn't mind an error or warn log here since we don't expect this to happen

service_capacity_modeling/tools/instance_families.py
homatthew: codex generated something like this:
```
# instance_families.py
"i7i": {
    ...
    "cpu_turbo_all_ghz": 3.2,
    "cpu_freq_from": "nflx",
},
"i7ie": {
    ...
    "cpu_turbo_all_ghz": 3.2,
    "cpu_freq_from": "nflx",
},
"p6-b200": {
    ...
    "cpu_turbo_all_ghz": 3.2,
    "cpu_freq_from": "nflx",
}
```

And then consume that

service_capacity_modeling/tools/auto_shape.py
jolynch: I'd rather just not use nflx in general, and we're pushing AWS to fix the APIs, I see this as a short term gap to work around bugs in AWS APIs not a long term strategy. Let's re-evaluate if the APIs are still wrong after our escalation resolves?

homatthew: If there's some conversation happening already to fix the APIs, then let's go down that route.

homatthew (approved):
Non blocking comments

hsane-dev (approved):

service_capacity_modeling/tools/instance_families.py
jolynch: Let's wait and see if AWS fixes their APIs https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/282#discussion_r3382022653

PR #283, opened 2026-06-05
Hsane/add 8a turin instances

hsane-dev:
## Title
Add AWS 8th gen AMD Turin instance families (m8a, r8a, c8a)

## Descrption

This adds support for AWS 8th gen AMD Turin instances (m8a, r8a, c8a) to SCM:
1. TURIN_IPC = 1.325 constant and three new instance family entries in instance_families.py, plus the generated shape files for all sizes of m8a, r8a, and c8a.
2. FREQUENCY = ALthough max frequency is 4.5GHz, under load we see degradation upto 3.6GHz. Current conservative value used is 3.8GHz for all core-turbo.

## Performance reasoning

Based on an m8a preview evaluation across 9 Netflix production services, Turin delivers roughly 25% CPU reduction versus m7a (Genoa). Almost all of that comes from frequency — the clock jumped from 3.7 GHz to 4.5 GHz (+21.6%). Architectural IPC improvement is minimal and mixed, with some regressions on compression and Java benchmarks (Renaissance). The 2% IPC uplift over Genoa (1.2995 → 1.325) reflects the midpoint of what the benchmarks show once you strip out the frequency contribution.

## Dependency note

This branch sits on top of hsane/fix-ht-ipc-scale-factor. Once that fix lands on main, this can be rebased cleanly. Without the fix, the IPC scale for Turin would be 1.325 instead of the correct 1.988.


hsane-dev:
@jolynch can you review addition of Turin (AMD 8th gen instances?)

baoalvin1 (approved):

service_capacity_modeling/hardware/profiles/shapes/aws/auto_m8a.json
```diff
@@ -0,0 +1,104 @@
+{
+  "instances": {
+    "m8a.medium": {
+      "name": "m8a.medium",
+      "cpu": 1,
+      "cpu_cores": 1,
+      "cpu_ghz": 4.5,
+      "cpu_ipc_scale": 1.98,
```
jolynch, line 8: This seems a bit high? Are we sure clock frequency is going to 4.5 _and_ ipc is going this high?

service_capacity_modeling/tools/instance_families.py
```diff
@@ -77,6 +78,13 @@
         "cpu_turbo_single_ghz": 3.8,
         "cpu_turbo_all_ghz": 3.2,
     },
+    "c8a": {
+        "iops_per_gib": None,
+        "io_latency_curve": None,
+        "cpu_ipc_scale": TURIN_IPC,
+        "cpu_turbo_single_ghz": 4.5,
```
jolynch, line 85: The EC2 apis appear broken again, they're reporting the turbo 4.5 as the sustained clock. 

```
aws --region us-east-1 ec2 describe-instance-types --instance-types m8a.4xlarge --output json | jq '.InstanceTypes[0].ProcessorInfo'
{
  "SupportedArchitectures": [
    "x86_64"
  ],
  "SustainedClockSpeedInGhz": 4.5,
  "Manufacturer": "AMD"
}
```

Pretty sure that should be 4 not 4.5. Let's reach out to AWS to fix their API and in the meantime use the flag on auto_shape to use our number (4.0)

hsane-dev: 4.5 is correct

service_capacity_modeling/hardware/profiles/shapes/aws/auto_m8a.json
baoalvin1: updated to use 4.0 cpu_ghz

service_capacity_modeling/tools/instance_families.py
jolynch: For single core yes,not for sustained all core.

jolynch (approved):
Switch to 4 looks good.

PR #271, opened 2026-04-21
fix: recalculate attached-drive sizing after count changes

homatthew:
## What am I trying to do?
Split the attached-drive replanning behavior change out of PR 265 into its own reviewable change.

## Why did I do it this way?
PR 265 should stay a transparent refactor for typed node-count explainability. The attached-drive iterative recalculation changes planner outputs and baseline costs, so it belongs in a separate stacked PR.

## Are there any tests?
Yes.
- `pytest tests/test_resource_counts.py -q`
- `pytest tests/netflix/test_cost_regression.py -k cassandra_kv_dense_ebs -q`

## How would I use the new code?
No API change. This changes planner behavior for attached-drive sizing when node count increases due to capacity or IOPS constraints.

PR #287, opened 2026-06-23
Fix attached drive sizing convergence

homatthew:
## What am I trying to do?

Fix attached-drive sizing so node count, per-node disk size, and per-node IOPS are recalculated until they agree. This is a fresh rewrite of PR #271 on latest `upstream/main`.

## Why did I do it this way?

The original loop direction was right, but the review found two edge cases: historical max counts could preserve transient IOPS overprovisioning, and a non-convergent loop could return stale drive IOPS. This version makes each sizing pass explicit, finds a stable upper count, tightens back to the smallest topology-valid count that still satisfies recomputed sizing, and converts attached-drive non-convergence into a planner `Excuse` instead of aborting the whole plan.

The old generated baseline diff no longer applies to current `main`; after running `tox -e capture-baseline`, no generated baseline files remain changed.

## Are there any tests?

- `tox -e py312 -- tests/test_resource_counts.py tests/test_generate_scenarios.py tests/netflix/test_cost_regression.py -k 'resource_counts or generate_scenarios or attached_drive_sizing_error or cassandra_kv_dense_ebs' -q`
- `tox -e capture-baseline`
- `tox -e pre-commit`

No API change. Existing callers of `compute_stateful_zone` get attached-drive plans whose returned drive size/IOPS and node-count context are based on the same converged node count.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -51,6 +51,7 @@ from service_capacity_modeling.interface import UncertainCapacityPlan
 from service_capacity_modeling.interface import ZoneClusterCapacity
 from service_capacity_modeling.models import CapacityModel
 from service_capacity_modeling.models import CostAwareModel
+from service_capacity_modeling.models.common import AttachedDriveSizingError
 from service_capacity_modeling.models.common import get_disk_size_gib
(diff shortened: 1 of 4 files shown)


homatthew:
Oops didn't mean to open tihs

PR #292, opened 2026-06-29
Cap Cassandra provisioned disk utilization

homatthew:
## What am I trying to do?

Cap Cassandra provisioned disk utilization at 55% after all storage and disk buffers are applied. This keeps additional headroom in generated plans while preserving the existing buffer composition model.

### Apply the cap as a final disk-buffer floor

Cassandra already has several storage/disk influences: model storage buffers, explicit caller buffers, and the EBS-hotter multiplier. The helper reads the current composite disk buffer and injects only the missing multiplicative factor needed to reach `1 / max_disk_utilization`.

That means:

- existing buffers above the cap are preserved
- buffers below the cap are raised to the cap
- EBS-hotter still runs first, but cannot push the final provisioned utilization above the configured cap

### Keep the limit stricter-only

The extra model argument is `max_disk_utilization`. It defaults to `0.55`, and validation only allows stricter values. The field exists so a workload can request more headroom, not less.

### Keep the estimator contract narrow

The estimator receives `max_disk_utilization` as a scalar rather than the full argument object. That keeps this policy consistent with the surrounding scalar arguments like `max_page_cache_gib` and `backup_retention_days`.

## Are there any tests?

Yes. The tests cover the caller-visible EBS sizing behavior, stricter utilization overrides, the default-buffer fallback edge in the cap helper, and validation that rejects looser or invalid utilization values.

Validation run:

```bash
tox -e py312 -- tests/netflix/test_cassandra_ebs_buffer.py
tox -e capture-baseline
tox -e pre-commit
```

The commit hook also reran the repo pre-commit path, including baseline capture.

Default planning now enforces 55% maximum provisioned Cassandra disk utilization. A caller can request stricter headroom through extra model arguments:

```json
{
  "max_disk_utilization": 0.50
}
```

The generated baselines show the expected impact: affected EBS Cassandra scenarios provision larger volumes and their annual cost increases accordingly.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/cassandra.py
+++ b/service_capacity_modeling/models/org/netflix/cassandra.py
@@ -78,10 +78,43 @@ logger = logging.getLogger(__name__)
 
 BACKGROUND_BUFFER = "background"
 EBS_HOTTER = "EBS_HOTTER"
+CASSANDRA_DISK_UTILIZATION_LIMIT = "CASSANDRA_DISK_UTILIZATION_LIMIT"
 CRITICAL_TIERS: Set[int] = {0, 1}
 # cluster size aka nodes per ASG
 CRITICAL_TIER_MIN_CLUSTER_SIZE = 2
 EBS_HOTTER_BUFFER_RATIO = 0.5
+CASSANDRA_MAX_DISK_UTILIZATION = 0.55
+
+
+def _with_disk_utilization_buffer(
+    desires: CapacityDesires,
+    max_disk_utilization: float,
+) -> CapacityDesires:
+    """Add only the disk buffer needed to keep Cassandra below the utilization cap.
+
+    The default target leaves margin below the previous cap. Callers may make
+    this stricter but not looser.
+    """
+    capped_desires = desires.model_copy(deep=True)
+    target_disk_buffer_ratio = 1 / max_disk_utilization
+    existing_disk_buffer = buffer_for_components(
+        buffers=desires.buffers,
+        components=[BufferComponent.disk],
+    )
+    disk_utilization_ratio = (
+        target_disk_buffer_ratio
+        if not existing_disk_buffer.sources
+        else max(1.0, target_disk_buffer_ratio / existing_disk_buffer.ratio)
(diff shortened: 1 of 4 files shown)


homatthew:
## Reviewer guide

Holistic read against `main`: this PR is a Cassandra disk-safety policy change, not a generic planner refactor.

Suggested review order:

1. Start with `service_capacity_modeling/models/org/netflix/cassandra.py` and check the new `_with_disk_utilization_buffer` helper. The key behavior is that it treats the disk utilization cap as a floor on the final composite disk buffer: raise too-low buffers, preserve already-larger buffers.
2. Then read `tests/netflix/test_cassandra_ebs_buffer.py`. The tests cover the EBS behavior, stricter overrides, the default-buffer fallback edge, and validation for invalid/looser `max_disk_utilization` values.
3. Finally skim the generated baselines. The expected behavioral evidence is larger EBS volumes and higher annual cost in affected Cassandra scenarios.

Why the implementation is shaped this way: Cassandra already composes storage, disk, and EBS-hotter buffers multiplicatively. Applying the utilization cap as one final disk buffer preserves that existing composition model while preventing the EBS-hotter optimization from planning above the configured utilization target.

homatthew:
Holistic note on the high-looking `effective_disk_per_node_gib` in the EBS baseline:

This output is expected for the existing-cluster path. The value is provisioned disk per node needed to keep the current dense EBS topology eligible, not a new-cluster data density target. For current EBS clusters, the model reads the observed disk utilization per node and current volume size, lifts the generic attached-data guardrail to at least the observed per-node data, then applies the disk utilization buffer to compute an effective provisioned-disk floor.

For the anonymized baseline, that means roughly `3000 GiB observed data/node * (1 / 0.55) ~= 5455 GiB`, rounded to `5500 GiB`. The emitted gp3 volume remains at the preserved existing floor (`5600GB`). That is also why `disk_capacity` can report `63` while `min_count` and final count remain `64`: the model is preserving a running dense EBS topology rather than forcing a wider cluster solely to satisfy the generic attached-disk guardrail.

PR #300, opened 2026-07-22
EVCache: soft-penalize instance-family migrations (different_family_regret)

shy-1234:
## Summary
Adds `different_family_regret` to `NflxEVCacheCapacityModel` — a soft penalty
that biases recommendations toward the instance family a cluster is already
running on, without excluding alternatives. **Defaults to `0.05`** (a 5%
anti-thrashing floor); set to `0` to disable.

## Motivation
When re-planning an existing EVCache cluster, the planner ranks purely by cost,
so it can recommend migrating to a *different* instance family that ranks only
marginally cheaper. Reservations are family-specific, so a family migration
carries real switching cost (on-demand premium until new reservations land) —
it isn't worth it for a couple percent of savings. A small default penalty
prevents this migration thrashing: don't switch families for <5% compute
savings.

This mirrors mechanisms already in the sibling models — Cassandra's
`different_family_regret` (also a soft rank penalty) and Kafka's
`require_same_instance_family` — and follows Cassandra's soft-penalty design.

## Behavior
- When `different_family_regret > 0` and a current EVCache cluster is present,
  a candidate whose `instance.family` differs from the current cluster's family
  gets `different_family_regret * compute_cost` added to its `plan.rank`
  (recorded under `RANK_PENALTIES`).
- **Soft, not hard:** a different family still wins when it is meaningfully
  cheaper, or when the current family can't satisfy the requirement — a plan is
  never returned empty for lack of same-family candidates.
- `total_annual_cost` stays honest (only `rank` is adjusted). The penalty
  applies to compute cost only, not fixed service (network) costs. `regret()`
  is unaffected — family migration is a fixed switching cost, so it biases the
  deterministic rank only (matching Cassandra's rationale).
- Only applies when the current EVCache cluster can be identified (see
  Composability below); no effect on greenfield planning (no current cluster).

## Composability (EVCache as a sub-model)
EVCache is composed by `key-value` / `read-only-kv` / `graphkv`, where
`current_clusters` may also carry the Cassandra tier — so `zonal[0]` is not
reliably the EVCache cluster. The new `_current_evcache_cluster` selects the
current cluster the penalty keys off:
- prefer a zonal entry tagged `cluster_type == "evcache"`;
- else fall back to `zonal[0]` **only** when there is a single current cluster
  (unambiguous, direct EVCache planning);
- else return `None` and skip the penalty (can't identify our tier → don't guess).

This keeps composite plans from being biased toward the Cassandra family.

## Tests
- `test_evcache_different_family_regret`: default applies the penalty (incumbent
  carries none; a different family carries one that inflates `rank` above the
  honest `total_annual_cost`; alternatives remain available); `0.0` disables it.
- `test_evcache_different_family_regret_flips_selection`: a cheaper family wins
  with the penalty off, and a large penalty flips the top pick back to the
  incumbent — proving the penalty changes the selected plan, not just annotates
  it.
- `test_evcache_family_penalty_targets_evcache_typed_current_cluster` and
  `test_evcache_family_penalty_skipped_for_ambiguous_untyped_current`: verify
  the composability rules above.

## Verification
- `pytest tests/netflix/` full suite passes; composed `key-value` /
  `read-only-kv` / `graphkv` suites unchanged.
- `ruff` (check + format), `flake8`, `mypy` (strict), `pylint` (10/10, no new
  waivers) all clean; `tox -e capture-baseline` produces no diff

### DIFF
--- a/service_capacity_modeling/models/org/netflix/evcache.py
+++ b/service_capacity_modeling/models/org/netflix/evcache.py
@@ -226,6 +226,7 @@ def _estimate_evcache_cluster_zonal(  # noqa: C901,E501 pylint: disable=too-many
     max_regional_size: int = 10000,
     min_instance_memory_gib: int = 12,
     cross_region_replication: Replication = Replication.none,
+    different_family_regret: float = 0.0,
 ) -> Union[CapacityPlan, Excuse, None]:
     # EVCache doesn't like to deploy on single CPU instances
     if instance.cpu < 2:
@@ -353,13 +354,11 @@ def _estimate_evcache_cluster_zonal(  # noqa: C901,E501 pylint: disable=too-many
     }
     upsert_params(cluster, params)
 
-    # Penalize under-spread clusters via rank (deterministic sort) and
-    # regret (stochastic robustness). Uses the same additive dollar formula
-    # as before, but applied to plan.rank instead of annual_costs — keeping
-    # total_annual_cost honest while preserving identical selection order.
+    # Penalize under-spread clusters via rank (deterministic sort) and regret
+    # (stochastic robustness). Additive dollar formula applied to plan.rank
+    # instead of annual_costs — keeping total_annual_cost honest. Combined with
+    # the family-migration penalty into RANK_PENALTIES below.
     spread_cost = calculate_spread_cost(cluster.count)
-    if spread_cost > 0:
-        upsert_params(cluster, {RANK_PENALTIES: {"under_spread": spread_cost}})
 
     # evcache clusters generally should try to stay under some total number
     # of nodes. Orgs do this for all kinds of reasons such as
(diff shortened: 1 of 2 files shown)


service_capacity_modeling/models/org/netflix/evcache.py
```diff
@@ -447,6 +503,20 @@ class NflxEVCacheArguments(BaseModel):
             "By default we do no replication"
         ),
     )
+    different_family_regret: float = Field(
+        default=0.0,
```
jolynch: This seems like a generally useful idea to prevent migration thrashing, since this only affects rightsizing any reason not to default this to say 5%? If we're not getting at least 5% ROI we shouldn't imo be migrating.

jolynch:
lg2m

service_capacity_modeling/models/org/netflix/evcache.py
shy-1234: The only reason was for backward compatibility. However, I think the use case is scoped (only when current_clusters is passed in), so should be fine to set the default to 5%.

shy-1234: One caveat: EVCache is also a sub-model of KV/graphkv, where the current cluster passed to the sub-model can be the Cassandra tier — so I scoped the penalty to key off the EVCache tier via cluster_type (falling back to the sole current cluster for direct planning, and skipping when it can't be identified)

PR #301, opened 2026-07-28
Stop tier-local reservations crossing model composition boundaries

homatthew:
## Why

A tier-0 KeyValue plan crashed with `ZeroDivisionError` in `compute_stateful_zone`:

```
File "service_capacity_modeling/models/common.py", line 563, in compute_stateful_zone
    needed_memory_gib / (instance.ram_gib - reserve_memory(instance.ram_gib))
ZeroDivisionError: float division by zero
```

The divide-by-zero turned out to be the visible edge of a class of bug: **things that describe one tier were crossing into another**. Each commit below fixes one instance of that, smallest first.

Draft because commits 3 and 8 change plan output for composed models, and the semantics deserve a second opinion before this is real.

## The chain that produced the crash

`reserved_instance_app_mem_gib: 24.0` is in the request. It comes from `dgi-artifact/deploys/kv/csticketing/prod.yaml`, hand-added in Dec 2024 ("Fixing the app memory") and carried forward untouched by the DABP bot ever since. It sizes the **dgwkv Java app's** heap.

`key_value.compose_with` handed desires to Cassandra with an identity lambda, so Cassandra reserved 24 GiB per node for an app that is not on the box. On `i4i.xlarge`:

```
ram 30.52  -  heap 15.0  -  reserves 25.0  =  -9.48  ->  MemoryLayout clamps to 0
reserve_memory(30.52) = 30.52 - 0 = 30.52
denominator = 30.52 - 30.52 = 0
```

47 shapes hit exactly zero at that reserve. Elasticsearch has no such clamp, so it went **negative** instead — and `math.ceil` of a negative quotient is a negative node count that `max()` silently discards, dropping the memory constraint entirely:

```
r5.xlarge, needed_memory_gib=1000, reserve = 3 + max(32, 15.26)
-> required_nodes_by_type = {'cpu': 1, 'memory': -287, ...}, bottleneck 'cpu', count: 1
```

One 30.52 GiB node "satisfying" a 1000 GiB requirement. That is why the fix is not a `max(1, ...)` on the denominator — ceiling it converts the crash you can see into the undersized cluster you cannot.

## Commits

| | commit | what |
|---|---|---|
| 1 | `7ed071b` | Reject shapes whose reservations leave no usable memory |
| 2 | `204a839` | Keep the KeyValue app memory reserve on the KeyValue tier |
| 3 | `2d3f0a1` | Let a model decide what crosses into the models it composes with |
| 4 | `12e8665` | Explain why Elasticsearch turns a shape down |
| 5 | `17b31fc` | Add a selector for a model's own entry in `current_clusters` |
| 6 | `9148142` | Read Cassandra's own current cluster, not whichever came first |
| 7 | `402ff06` | Screen shapes against the model's own current cluster |
| 8 | `404ca11` | Compose desires down a chain, not from the original request |

**1** — `compute_stateful_zone` raises instead of guessing when per-node usable memory is `<= 0`; Cassandra/Kafka/EVCache/Elasticsearch reject the shape first. Cassandra excuses with the arithmetic in the reason.

**2–3** — `reserved_instance_app_mem_gib` is memory per instance of the tier it was written for. `CapacityModel.desires_for_composed_model` is a new overridable hook the planner applies to everything a model composes with; the default drops the reserve so the composed model's own `default_desires` supplies it. Commit 2 fixes KeyValue narrowly and ships on its own; commit 3 generalises and removes the special case. It runs *before* the per-child transform, so a model with a reason to pass its reserve down can still say so.

**4** — Elasticsearch returned bare `None` from all eight rejection paths while every other stateful model returned an `Excuse`. The base class already allowed it.

**5–7** — `CurrentClusterCapacity.cluster_type` is documented as "required metadata for identifying which model this capacity belongs to" and nothing read it; every consumer took `zonal[0]`. Cassandra and EVCache are both zonal, so a composed request could hand Cassandra EVCache's node count, disk and **instance shape** (via `CapacityDesires.reference_shape`, which also reads `zonal[0]` — Cassandra was normalizing its core counts against a cache node). Unlabelled entries fall back to the first, so existing callers are unaffected.

**8** — `compose_with` transforms received the original user desires, so only the first level of a chain took effect. GraphKV → KeyValue → Cassandra dropped GraphKV's amplification: Cassandra was sized for 5k logical reads/sec when it serves 275k.

## Plan impact

24 GiB reserve, memory-bound namespace, per zone:

| model | before | after |
|---|---|---|
| key-value, time-series | `cassandra i4i.2xlarge` | `i4i.xlarge` (−21%) |
| entity | | −14% |
| graphkv | `i4i.xlarge x2` | `c6id.4xlarge x16` (fan-out finally provisioned for) |
| elasticsearch data nodes | `i4i.xlarge x9` | `i4i.2xlarge x5` (dropped constraint applied) |
| cassandra, kafka, evcache, wal, control (direct) | unchanged | unchanged |

Directly planning a model still applies the caller's reserve to it — there is a test for that.

## What I deliberately did not do

- **No override of `desires_for_composed_model` anywhere.** `org.netflix.elasticsearch` looked like a candidate (it owns no instances, only splits into node roles) but it is never planned directly — dgi-artifact names nine models and that is not one of them. Entity is the only composition reaching Elasticsearch and it strips a level higher anyway, so an override changes nothing except adding a second mode. Verified by diffing plan output with and without it.
- **`reserved_instance_system_mem_gib` still propagates.** Kernel/OS overhead is roughly shape-independent and no model overrides the default, so stripping it would change nothing today while removing the caller's ability to state it.
- **`elasticsearch.py` reserves `base_mem + max(32, ram / 2)`** where its own comment says "half of available system max of 32 for compressed oops" — i.e. `min`. Very likely a bug, but flipping it is an ES heap-sizing decision for whoever owns that model. Commit 1 rejects the shapes it overruns rather than changing the formula.
- **`key_value.compose_with` divides `rps / wps`** on unmerged desires, so a read-only KeyValue request crashes before planning starts. Found while tracing this; separate bug, untouched.

## Testing

782 passed, 1 skipped. mypy clean, pylint 10.00, cost baselines unchanged. Every new test verified to fail against the pre-fix code.

`test_all_models_tier_capacity_relationship` is flaky on `main` independently of this branch — it draws random workloads and some violate the property. Three different falsifying draws (R:47480/W:19623, R:50000/W:14583, R:49307/W:15969, all at 100 GiB) produce byte-identical tier-0-worse-than-tier-2 numbers on `main` and on this branch.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -1488,10 +1488,16 @@ class CapacityPlanner:
             )
 
             # We might have to compose this model with others depending on
-            # the user requirement
+            # the user requirement. What crosses that boundary is the composing
+            # model's call -- see CapacityModel.desires_for_composed_model,
+            # which by default drops reservations describing this model's own
+            # instances. It runs before the per-child transform so a transform
+            # that sets something deliberately still wins.
+            child_desires = self._models[sub_model].desires_for_composed_model(desires)
+
             queue.extend(
                 [
-                    (modify_child_desires(desires), child_model)
+                    (modify_child_desires(child_desires), child_model)
                     for child_model, modify_child_desires in self._models[
                         sub_model
                     ].compose_with(desires, extra_model_arguments)
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -1493,14 +1493,21 @@ class CapacityPlanner:
             # which by default drops reservations describing this model's own
             # instances. It runs before the per-child transform so a transform
             # that sets something deliberately still wins.
-            child_desires = self._models[sub_model].desires_for_composed_model(desires)
+            #
+            # Both start from parent_desires, the desires this model was handed,
+            # so a chain composes: GraphKV amplifies logical traffic into KeyValue
+            # operations, and Cassandra sees the amplified figures rather than the
+            # user's original ones.
+            child_desires = self._models[sub_model].desires_for_composed_model(
+                parent_desires
+            )
 
             queue.extend(
                 [
                     (modify_child_desires(child_desires), child_model)
                     for child_model, modify_child_desires in self._models[
                         sub_model
-                    ].compose_with(desires, extra_model_arguments)
+                    ].compose_with(parent_desires, extra_model_arguments)
                 ]
             )
 
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -61,6 +61,7 @@ from service_capacity_modeling.interface import SampleRef
 from service_capacity_modeling.interface import ZoneClusterCapacity
 from service_capacity_modeling.models import CapacityModel
 from service_capacity_modeling.models import CostAwareModel
+from service_capacity_modeling.models.common import current_cluster_capacity
 from service_capacity_modeling.models.common import get_disk_size_gib
 from service_capacity_modeling.models.common import merge_plan
 from service_capacity_modeling.models.org import netflix
@@ -1064,7 +1065,9 @@ class CapacityPlanner:
         )
 
     # Calculates the minimum cpu, memory, and network requirements based on desires.
-    def _per_instance_requirements(self, desires: CapacityDesires) -> Tuple[int, float]:
+    def _per_instance_requirements(
+        self, model: CapacityModel, desires: CapacityDesires
+    ) -> Tuple[int, float]:
         # Applications often set fixed reservations of heap or OS memory
         per_instance_mem = (
             desires.data_shape.reserved_instance_app_mem_gib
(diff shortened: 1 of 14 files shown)


homatthew:
Superseded by a stacked split so the memory fix can land on its own:

- #303 - Stop an app memory reserve meant for one tier from sizing another (the reported crash)
- #304 - Explain why Elasticsearch turns a shape down
- #305 - Read the current cluster that belongs to this model
- #306 - Compose desires down a chain, not from the original request
- #308 - Report the desires a composed model was actually planned with

The split also carries fixes this branch predates: a partially-labelled
`current_clusters` payload discarded a model's own unlabelled entry (a verified
3.5x Cassandra under-provision), `cluster_type` was matched against 16 models that
never declared it, and inheritance is now configured per composition edge via
ChildDesiresConfig.

PR #302, opened 2026-07-28
fix(models): preserve memory boundaries in composed plans

homatthew:
## What am I trying to do?

Fix the KeyValue planner crash caused when a parent application's per-instance memory reservation reaches composed Cassandra nodes and leaves no usable memory. The planner should reject infeasible instance shapes cleanly, while composition preserves workload transforms such as GraphKV fanout without applying one service's app-memory reservation to another service's hosts.

This is a draft implementation for review because `compose_with` currently serves both service composition and Elasticsearch role expansion.

### Configure inheritance on each parent-to-child edge

`ChildDesiresConfig` is a typed Pydantic policy exposed by the base `CapacityModel`. Its narrow initial setting controls whether `reserved_instance_app_mem_gib` crosses a specific composition edge.

The default resets that field before the existing edge transform runs. This lets Cassandra and EVCache receive their own model defaults instead of KeyValue's app reservation. Elasticsearch explicitly opts its data, master, and search roles into inheritance because those children are roles of the same service rather than separately hosted dependencies.

The edge transform runs after the reset, so a model can still deliberately assign a child-specific reservation.

### Preserve workload transformations through multi-level composition

Composition now starts each child from desires already transformed by its ancestors. For example, GraphKV's read/write fanout reaches KeyValue and then Cassandra; parent model defaults remain local and do not leak into children.

```mermaid
flowchart LR
    A[Caller workload] --> B[Graph service fanout]
    B --> C[Key/value workload]
    C --> D[Cassandra workload]
    C --> E[Cache workload]
```

Read/write rates and other caller workload settings continue through the chain. Only the parent app-memory reservation is reset by default.

### Reject impossible memory shapes explicitly

Shared stateful sizing now refuses shapes whose reservations leave zero or negative usable memory. Cassandra and Elasticsearch reject those candidates with structured memory `Excuse` values, allowing the planner to continue evaluating feasible shapes instead of dividing by zero or silently dropping the constraint.

### Route current deployment state to its owner

When composed requests contain typed `current_clusters`, each model receives only cluster types it owns. Untyped rows remain a fallback during partial adoption so existing callers are not cut over abruptly.

## Alternatives considered

- **Special-case KeyValue:** narrow for the reported request, but leaves the same ambiguous inheritance behavior at every other composition edge.
- **Strip app memory globally with an “owns a tier” flag:** cannot accurately represent Elasticsearch's role expansion and makes the policy an indirect property of the parent rather than an explicit child-edge decision.
- **Pass either all desires or no desires:** breaks legitimate propagation such as GraphKV → KeyValue → Cassandra RPS/WPS fanout.
- **Hardcode Cassandra or EVCache reserve values in the parent:** duplicates child defaults in the wrong model and would drift as those defaults change.
- **Strip additional fields now:** system memory, buffers, parallelism, and latency semantics are less clearly service-local. This PR leaves them unchanged rather than expanding the behavioral contract without evidence.
- **Change the existing `compose_with` tuple shape or add decorators:** creates broader API churn. A base-model policy hook keeps existing composition implementations compatible and allows per-child overrides.

## Are there any tests?

Yes. Regression coverage verifies the original 24 GiB request no longer crashes, infeasible shapes produce memory excuses, KeyValue's reservation does not reach Cassandra or EVCache, Elasticsearch role models explicitly inherit it, multi-level RPS/WPS transforms survive, and typed/untyped current-cluster routing remains compatible.

The no-mistakes gate passed focused tests and lint. GitHub Actions passed the end-to-end suite and linters on Python 3.10, 3.11, and 3.12.

Most composed-service edges use the safe default:

```python
ChildDesiresConfig()
```

A same-service role expansion can explicitly inherit the reservation:

```python
@staticmethod
def child_desires_config(child_model: str) -> ChildDesiresConfig:
    return ChildDesiresConfig(inherit_app_memory_reservation=True)
```

## Deliberately deferred

`PlanExplanation.desires_by_model` currently reconstructs composed children from root desires, so its explanation can disagree with the accumulated transforms/defaults used by planning. The no-mistakes review caught this; it is fixed separately in #307 so this behavioral fix remains reviewable.

Changes to dgi-artifact inputs and broader desire-field inheritance semantics are also outside this PR.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -1556,16 +1556,16 @@ class CapacityPlanner:
             # exception: the caller was describing the composed models all
             # along. Strip before the modifier runs so a model that
             # deliberately sets a child's reservation still wins.
-            child_desires = desires
+            child_desires = parent_desires
             if self._models[sub_model].plans_own_cluster():
-                child_desires = _without_app_memory_reserve(desires)
+                child_desires = _without_app_memory_reserve(parent_desires)
 
             queue.extend(
                 [
                     (modify_child_desires(child_desires), child_model)
                     for child_model, modify_child_desires in self._models[
                         sub_model
-                    ].compose_with(desires, extra_model_arguments)
+                    ].compose_with(parent_desires, extra_model_arguments)
                 ]
             )
 
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -503,6 +503,21 @@ def _resolve_cluster_instances(desires: CapacityDesires) -> None:
                 cap.cluster_instance = shapes.instance(cap.cluster_instance_name)
 
 
+def _without_app_memory_reserve(desires: CapacityDesires) -> CapacityDesires:
+    """Drop the app memory reserve so a composed model uses its own.
+
+    ``reserved_instance_app_mem_gib`` is memory per instance of the tier it
+    was written for, and composed models run on their own shapes. A KeyValue
+    request that reserves 24 GiB for the dgwkv Java app should not make
+    Cassandra reserve 24 GiB per node for an app that is not there. Unsetting
+    the field (rather than picking a number) lets each model's
+    ``default_desires`` supply the reserve that suits it.
+    """
+    data_shape = desires.data_shape.model_dump(exclude_unset=True)
+    data_shape.pop("reserved_instance_app_mem_gib", None)
+    return desires.model_copy(deep=True, update={"data_shape": DataShape(**data_shape)})
+
+
 class CapacityPlanner:
     def __init__(
         self,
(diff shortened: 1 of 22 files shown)


homatthew:
Superseded by a stacked split so the memory fix can land on its own:

- #303 - Stop an app memory reserve meant for one tier from sizing another (the reported crash)
- #304 - Explain why Elasticsearch turns a shape down
- #305 - Read the current cluster that belongs to this model
- #306 - Compose desires down a chain, not from the original request
- #308 - Report the desires a composed model was actually planned with

The split also carries fixes this branch predates: a partially-labelled
`current_clusters` payload discarded a model's own unlabelled entry (a verified
3.5x Cassandra under-provision), `cluster_type` was matched against 16 models that
never declared it, and inheritance is now configured per composition edge via
ChildDesiresConfig.

PR #304, opened 2026-07-28
Explain why Elasticsearch turns a shape down

homatthew:
**PR 2 of 4**, stacked on #303. Review #303 first — this only makes sense on top of it.

## Why

Cassandra, Kafka and EVCache each return an `Excuse` naming the reason and the bottleneck
when they turn an instance down, and the planner collects those into
`plan_certain_explained().excuses_by_model`. Elasticsearch returned a bare `None` from all
eight of its rejection paths, so a caller asking why their instance family never appeared
had nothing to read.

This is the "confusing provisioning" case: a shape silently vanishes from consideration and
the plan gives no account of it.

The base `CapacityModel.capacity_plan` already declares
`Union[CapacityPlan, Excuse, None]` and `_plan_certain` already has a `case Excuse()` arm.
Elasticsearch had simply narrowed its own annotation to `Optional[CapacityPlan]` and never
adopted the mechanism.

## What changed

The data, master and search node models now excuse instead of returning `None`:

| bottleneck | reason |
|---|---|
| `cpu` / `memory` | `Instance too small: 2 vCPUs (min 2), 15 GiB RAM (min 24 GiB)` |
| `memory` | `Reserved memory (35 GiB) leaves no RAM for shards on a 29 GiB shape` |
| `drive_type` | `Requires local disks but m6a.4xlarge is EBS-only` |
| `cluster_size` | `Needs 41 nodes per zone, over the 40 allowed by max_regional_size (120)` |

The aggregator's `capacity_plan` keeps returning `None` — that one is not a rejection, it
owns no instances at all and exists only to split the service into node roles.

Two mechanical tidies came with it: `max_zonal_size` extracted from the inline
`max_regional_size // zones_in_region`, and the master/search `ram <= 24` and `cpu <= 2`
checks merged into one condition so the excuse can attribute the right bottleneck. Both are
logically identical to what they replace.

## Behavior

No plan changes. `Excuse` and `None` are treated identically for control flow in
`_plan_certain`, and the aggregator never invokes the node models directly, so there is no
`is None` check an `Excuse` could slip past. The only observable difference is that
`excuses_by_model` is now populated for Elasticsearch.

## Testing

Full suite green. `test_es_rejections_explain_themselves` asserts every excuse carries a
reason and a bottleneck, and that both `drive_type` and `memory` are represented — it fails
against the unfixed code.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/elasticsearch.py
+++ b/service_capacity_modeling/models/org/netflix/elasticsearch.py
@@ -5,6 +5,7 @@ from typing import Callable
 from typing import Dict
 from typing import Optional
 from typing import Tuple
+from typing import Union
 
 from pydantic import BaseModel
 from pydantic import Field
@@ -14,6 +15,7 @@ from service_capacity_modeling.interface import AccessPattern
 from service_capacity_modeling.interface import Buffer
 from service_capacity_modeling.interface import BufferComponent
 from service_capacity_modeling.interface import Buffers
+from service_capacity_modeling.interface import Bottleneck
 from service_capacity_modeling.interface import CapacityDesires
 from service_capacity_modeling.interface import CapacityPlan
 from service_capacity_modeling.interface import CapacityRequirement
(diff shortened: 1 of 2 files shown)


homatthew:

PR #305, opened 2026-07-28
Read the current cluster that belongs to this model

homatthew:
**PR 3 of 4**, stacked on #304. Review #303 and #304 first.

## Why

`CurrentClusterCapacity.cluster_type` carries this comment in `interface.py`:

```python
# Optional: if not set, extract_baseline_plan
# Required metadata for identifying which model
# this capacity belongs to
cluster_type: Optional[str] = None
```

Nothing read it. Every consumer took `current_clusters.zonal[0]` or `regional[0]`.

That is fine for a single-model request and wrong for a composed one. **Cassandra and
EVCache are both zonal**, so a KeyValue request describing both tiers put them in one list
and each model read whichever entry the caller happened to list first. Reorder the list and
the plan changes.

This is squarely a confusing-provisioning bug: Cassandra sized against a cache tier's node
count, disk utilization and instance shape, with nothing in the output saying so.

## Commits

| commit | what |
|---|---|
| `308d3e5` | Give every model that owns clusters a `cluster_type` |
| `ad158ee` | Add a selector for a model's own entry in `current_clusters` |
| `dbd7ef0` | Read Cassandra's own current cluster, not whichever came first |
| `76c9a48` | Screen shapes against the model's own current cluster |
| `0d85709` | Read EVCache's own current cluster too |

**`308d3e5` — the prerequisite.** For a model to find its own entry it has to know its own
name, and only six did. `cluster_type` was declared on the `CostAwareModel` mixin, where
its documented job is filtering which clusters a model *costs*; identifying your own
current cluster is a modeling concern, so it moves to `CapacityModel` with an empty
default. Thirteen models now declare the label they already stamp on the clusters they
produce — `dgwcontrol`, `dgwcounter`, `dgwentity`, `dgwgraphkv`, `dgwidentifier`, `dgwts`,
`dgwwal`, `aurora-cluster`, `rds-cluster`, `zk-zonal`, and the three Elasticsearch node
roles. No new names; every value is lifted from a literal already in the file.

Three keep the empty default: `org.netflix.elasticsearch` owns no clusters, which is what
empty means, and Postgres and DynamoDB stamp no label and read no current cluster.

**`ad158ee` — the selector.** `current_cluster_capacity(desires, cluster_type)` matches on
the label, then falls back to the first *unlabelled* entry, then returns `None`:

```
payload                      before this PR    with the selector
both labelled                evcache's         mine
partial: mine unlabelled     mine              mine
none labelled                first             first
only someone else's          evcache's         None
```

The fallback matters. An unlabelled entry means *nobody said whose it is*, not *it belongs
to someone else* — so it is still the best guess available, exactly as before. Only entries
labelled as another tier are ruled out. Getting that wrong is a 3.5× under-provision
(Cassandra 8 nodes → 2) the moment a caller labels one tier before another, which is what
a rollout looks like.

**`dbd7ef0`** — Cassandra read `zonal[0]` in three places: `_get_current_capacity` (cluster
size, derived buffers, family-change penalty), `_estimate_zonal_data_gib` (observed disk),
and its reference shape. That last one came from `CapacityDesires.reference_shape`, a
property with no model context that also reads `zonal[0]` — so Cassandra was normalizing
its core counts against a cache node. `zonal_requirements_from_current` now takes the
caller's chosen entry rather than the whole list, since picking one needs a `cluster_type`
the helper does not have.

**`76c9a48`** — `generate_scenarios` screens out shapes below a floor derived partly from
the current cluster's observed memory. That floor came from `zonal[0]` too, so one tier's
utilization could screen shapes out of another tier's search.

**`0d85709`** — the other half of the collision. Fixing only Cassandra left EVCache sizing
itself off Cassandra's cluster.

## Behavior

Unchanged unless the caller labels their current clusters with `cluster_type`, which is
what makes any of this reachable. Existing unlabelled payloads take the same path as
before.

## Known limitation, not addressed here

`CapacityDesires.reference_shape` raises `ValueError` when `current_clusters` holds both
zonal and regional entries. A composed KeyValue request has regional `dgwkv` plus zonal
`cassandra`/`evcache`, so **that full payload cannot be planned today** — pre-existing, and
why the zonal+zonal Cassandra/EVCache case is the one this PR fixes. Routing the two
remaining `desires.reference_shape` readers (`stateless_java.py`, `evcache.py`) through the
selector would lift that, and belongs in its own change.

## Testing

782 passed, 1 skipped; the one failure is `test_all_models_tier_capacity_relationship`,
which is flaky on `main` independently of this branch (three falsifying draws all give
byte-identical numbers on `main`). Both integration tests — Cassandra and EVCache each
ignoring the other's entry — fail without their respective fix.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -1101,7 +1101,7 @@ class CapacityPlanner:
         )
 
         current_capacity = current_cluster_capacity(
-            desires, getattr(model, "cluster_type", "")
+            desires, *model.current_cluster_types()
         )
         # Return early if we dont have current_capacity set.
         if current_capacity is None or current_capacity.cluster_instance is None:
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -62,6 +62,7 @@ from service_capacity_modeling.interface import ZoneClusterCapacity
 from service_capacity_modeling.models import CapacityModel
 from service_capacity_modeling.models import ChildDesiresConfig
 from service_capacity_modeling.models import CostAwareModel
+from service_capacity_modeling.models.common import current_cluster_capacity
 from service_capacity_modeling.models.common import get_disk_size_gib
 from service_capacity_modeling.models.common import merge_plan
 from service_capacity_modeling.models.org import netflix
@@ -1082,7 +1083,9 @@ class CapacityPlanner:
         )
 
     # Calculates the minimum cpu, memory, and network requirements based on desires.
-    def _per_instance_requirements(self, desires: CapacityDesires) -> Tuple[int, float]:
+    def _per_instance_requirements(
+        self, model: CapacityModel, desires: CapacityDesires
+    ) -> Tuple[int, float]:
         # Applications often set fixed reservations of heap or OS memory
         per_instance_mem = (
             desires.data_shape.reserved_instance_app_mem_gib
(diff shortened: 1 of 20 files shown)


homatthew:

PR #309, opened 2026-07-28
Cap the Elasticsearch heap at 32 GiB instead of flooring it there

homatthew:
**Stack:** PR 5 of 5, based on #306. This Elasticsearch fix does not depend on the planner behavior changed in the earlier layers.

## What am I trying to do?

Cap the Elasticsearch JVM heap at 32 GiB instead of treating 32 GiB as its minimum size.

The current calculation reserves `base_mem + max(32, ram / 2)`. That is backwards: Elasticsearch uses half of instance RAM for heap until the heap reaches 32 GiB, then keeps the 32 GiB cap so the JVM can continue using compressed object pointers.

## What changes?

The reservation becomes:

```python
heap_gib = min(32, instance.ram_gib / 2)
reserved_gib = app_and_system_reserves_gib + heap_gib
usable_gib = instance.ram_gib - reserved_gib
```

Assuming the existing 2 GiB app and system reserve, the corrected calculation works like this:

| Stored instance RAM | Heap before | Heap after | Usable RAM before | Usable RAM after |
|---:|---:|---:|---:|---:|
| 30.52 GiB | 32.00 GiB | 15.26 GiB | -3.48 GiB | 13.26 GiB |
| 61.04 GiB | 32.00 GiB | 30.52 GiB | 27.04 GiB | 28.52 GiB |
| 122.07 GiB | 61.04 GiB | 32.00 GiB | 59.03 GiB | 88.07 GiB |

The middle row explains why the bug was hard to see. On an advertised 64 GiB instance, the old formula happened to produce the expected 32 GiB heap. Smaller instances could end up with no RAM for shards and page cache, while larger instances reserved increasingly oversized heaps.

## What does this change in a capacity plan?

Callers do not need to change anything. Elasticsearch planning automatically evaluates instance shapes with the corrected heap reservation.

Examples from tier 1 data-node plans:

| Workload | Before | After | Annual cost change |
|---|---|---|---:|
| 200 GiB state, 10k RPS | `r5d.2xlarge x1` | `i4i.xlarge x1` | -20% |
| 1 TiB state, 10k RPS | `i3.2xlarge x1` | `i7ie.xlarge x1` | -15% |
| 32 TiB state, 20k RPS | `i3en.2xlarge x12` | `i3en.3xlarge x7` | -12% |

The 8 TiB / 50k RPS example remains `i3en.2xlarge x3`, so the correction does not force every plan onto a different shape.

## Why use a named constant?

`ES_MAX_HEAP_GIB = 32` makes the JVM boundary explicit and gives the test one source of truth.

`Instance.ram_gib` is stored about 4.6% below the advertised value for many AWS families. For example, an advertised 64 GiB shape is stored as 61.04 GiB, so its calculated heap is 30.52 GiB rather than 32 GiB. That stays below the compressed-object-pointer boundary. Correcting the shared RAM data would affect every capacity model and is intentionally outside this PR.

## Are there any tests?

Yes. The regression test verifies that every selected Elasticsearch data node retains RAM for shards and page cache after app, system, and heap reservations, and that the heap never exceeds 32 GiB.

CI passes on Python 3.10, 3.11, and 3.12. Elasticsearch is not present in the generated cost baselines, so this change does not regenerate baseline fixtures.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/elasticsearch.py
+++ b/service_capacity_modeling/models/org/netflix/elasticsearch.py
@@ -48,6 +48,17 @@ from service_capacity_modeling.stats import dist_for_interval
 
 logger = logging.getLogger(__name__)
 
+# The JVM uses compressed object pointers below a ~32 GiB heap. Cassandra caps
+# its own heap the same way (DEFAULT_MAX_HEAP_GIB in cassandra_memory).
+#
+# Note this is compared against Instance.ram_gib, which for many families is
+# stored ~4.6% below the advertised figure (an AWS-advertised 32 GiB instance is
+# 30.52 here: 32768 MiB read as MB and converted to GiB). So on a 64 GiB shape
+# the cap yields a 30.52 GiB heap rather than 32 -- under the boundary, which is
+# the safe side. Correcting ram_gib would move sizing for every model and is a
+# separate change.
+ES_MAX_HEAP_GIB = 32
+
 
 def _target_rf(desires: CapacityDesires, user_copies: Optional[int]) -> int:
     if user_copies is not None:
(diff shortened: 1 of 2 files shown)


homatthew:
superseded by #311

PR #311, opened 2026-07-29
fix(models): restore Elasticsearch memory reservation behavior after central fix

homatthew:
## Summary

#311 restores valid Elasticsearch planning after #303 exposed the existing heap calculation bug. Before #303, smaller shapes could have negative usable RAM; the resulting negative memory count was discarded when the planner combined CPU, disk, and memory requirements. #311 retains #303's nonpositive-memory guard and changes the heap reservation from a 32 GiB floor to a 32 GiB cap.

```python
old_reserve = app_memory + system_memory + max(32, instance_ram / 2)
new_reserve = app_memory + system_memory + min(32, instance_ram / 2)
usable = instance_ram - new_reserve
```

## Before #303 / after #311

The following results replay the same four tier-1 `org.netflix.elasticsearch` requests in `us-east-1` against the commit before #303 and the head of #311. Each request uses a state interval of `(0.1x, 1x, 4x)`, a read-rate interval of `(0.1x, 1x, 10x)`, and confidence `0.98`. No instance-family filter is applied.

| Midpoint workload | Before #303 data nodes per zone | After #311 data nodes per zone | Composed annual cost before / after | Change |
| --- | --- | --- | ---: | ---: |
| 200 GiB, 10k reads/s | `i4i.xlarge` x1, -4.48 GiB usable | `i4i.xlarge` x1, 12.26 GiB usable | $8,336.97 / $8,336.97 | 0% |
| 1,000 GiB, 10k reads/s | `i7ie.xlarge` x1, -4.48 GiB usable | `i7ie.xlarge` x1, 12.26 GiB usable | $9,880.50 / $9,880.50 | 0% |
| 8,000 GiB, 50k reads/s | `i7ie.xlarge` x5, -4.48 GiB usable | `i3en.2xlarge` x3, 27.52 GiB usable | $30,578.58 / $31,696.98 | +3.7% |
| 32,000 GiB, 20k reads/s | `i3en.xlarge` x19, -4.48 GiB usable | `i3en.3xlarge` x7, 56.55 GiB usable | $90,187.17 / $99,178.05 | +10.0% |

Costs include data, master, and search tiers. Node counts are per zone. These are planner replays, not measurements of production clusters.

Every before-#303 result in this table is invalid: the selected data shape has negative usable RAM. Its memory requirement was reported as 0 or a negative number, then ignored in favor of CPU, disk, or minimum-count requirements. After #311, each selected shape has positive usable RAM and a positive memory count. Disk still determines the final count for the 32,000 GiB workload; memory, CPU, and disk jointly determine it for the 8,000 GiB workload.

## Focused regression case

For a 100 GiB, 1,000 reads/s request restricted to `r5d`, both revisions select one `r5d.xlarge` per zone at $8,538.03 annually for the composed plan. The calculation changes from:

```text
Before #303: 31.7 - 2 app - 1 system - 32 heap = -3.30 GiB usable
After #311:  31.7 - 2 app - 1 system - 15.85 heap = 12.85 GiB usable
```

The memory count changes from 0 to 1. On a 128 GiB `r5d.4xlarge`, modeled usable memory changes from 61 GiB before #303 to 93 GiB after #311 because the heap is capped at 32 GiB.

When the caller omits app memory, data, master, and search roles use the generic 2 GiB default both before #303 and after #311. An explicit caller reservation still reaches all three roles. With a 16 GiB app reservation, the guard rejects `r5d.xlarge` at -1.15 GiB usable; `r5d.2xlarge` remains eligible at 14.81 GiB.

## Scope and verification

The change is limited to the Elasticsearch model and regression tests. It does not include the broader `Excuse` conversion or unrelated stacked changes. There are no Elasticsearch entries in the generated baseline fixtures, so the comparisons above come from direct planner replays at the two revisions.

Planner tests cover selected shape, cost, effective role reservation, usable memory, memory count, explicit reservation propagation, and rejection of shapes with nonpositive usable memory. The GitHub test matrix passed on Python 3.10, 3.11, and 3.12. The four-workload experiment was run twice at each revision with identical output.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/elasticsearch.py
+++ b/service_capacity_modeling/models/org/netflix/elasticsearch.py
@@ -571,9 +571,7 @@ class NflxElasticsearchCapacityModel(CapacityModel):
                         high=4,
                         confidence=0.98,
                     ),
-                    # Elasticsearch has a 1 GiB sidecar. This model plans no
-                    # instances, so the figure stops here -- see
-                    # child_desires_config for what reaches the node roles.
+                    # Elasticsearch has a 1 GiB sidecar
                     reserved_instance_app_mem_gib=1,
                 ),
             )
@@ -635,9 +633,7 @@ class NflxElasticsearchCapacityModel(CapacityModel):
                         high=4,
                         confidence=0.98,
                     ),
-                    # Elasticsearch has a 1 GiB sidecar. This model plans no
-                    # instances, so the figure stops here -- see
-                    # child_desires_config for what reaches the node roles.
+                    # Elasticsearch has a 1 GiB sidecar
                     reserved_instance_app_mem_gib=1,
                 ),
             )
--- a/service_capacity_modeling/models/org/netflix/elasticsearch.py
+++ b/service_capacity_modeling/models/org/netflix/elasticsearch.py
@@ -195,10 +195,7 @@ class NflxElasticsearchDataCapacityModel(CapacityModel):
     ) -> CapacityDesires:
         desires = CapacityModel.default_desires(user_desires, extra_model_arguments)
         desires.buffers = NflxElasticsearchDataCapacityModel.default_buffers()
-        # Elasticsearch has a 1 GiB sidecar. That figure was only stated on the
-        # aggregator model, which owns no instances, so the nodes that actually
-        # reserve the memory fell through to the DataShape default of 2. State
-        # it here, where the instances are.
+        # Elasticsearch data nodes run a 1 GiB sidecar.
         desires.data_shape.reserved_instance_app_mem_gib = 1
         return desires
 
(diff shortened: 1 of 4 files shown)


service_capacity_modeling/models/org/netflix/elasticsearch.py
```diff
@@ -195,11 +195,6 @@ def default_desires(
     ) -> CapacityDesires:
         desires = CapacityModel.default_desires(user_desires, extra_model_arguments)
         desires.buffers = NflxElasticsearchDataCapacityModel.default_buffers()
-        # Elasticsearch has a 1 GiB sidecar. That figure was only stated on the
-        # aggregator model, which owns no instances, so the nodes that actually
-        # reserve the memory fell through to the DataShape default of 2. State
-        # it here, where the instances are.
-        desires.data_shape.reserved_instance_app_mem_gib = 1
```
homatthew, line 202: This artifact came from mr PR in #303 where I took the reserved app mem gib from the aggregator model. It should be propagated from the parent. It wasn't necessary to set it here.

homatthew:
If it turns out we are over provisioning going forward, we can just disable the memory checks altogether to effectively go back to what it was before but I think it's better to fix forward

PR #314, opened 2026-08-05
feat(cassandra): attribute service cost per keyspace replication

homatthew:
## What am I trying to do?

Attribute Cassandra network and backup service costs to the keyspaces that actually incur them. A cluster can mix globally replicated keyspaces with region-confined keyspaces; pricing the entire plan as one placement charges confined data for cross-region traffic it never sends.

Existing callers remain source-compatible. When `keyspace_replication` is omitted, the established single-replication path is unchanged.

### Accept placements once, derive regional planner inputs

`KeyspacePlacement` is the caller-facing shape: keyspaces, copies per region, deployed regions, and relative state/write weights. `keyspace_replication_by_region()` converts those facts into `KeyspaceReplication` entries for each regional plan, deriving `num_regions` and regional scoping instead of asking every caller to reproduce that logic.

State weights are normalized among placements present in each region. Write weights are normalized once across the fleet because each regional plan carries the fleetwide write rate. The generated inputs therefore preserve additive fleet costs and do not repeat a single-region placement in regions where it does not live.

Direct `KeyspaceReplication` inputs remain validated: state shares must total approximately 0 or 1, write shares may total at most 1 within `1e-6`, and `copies_per_region` cannot be supplied alongside placements.

### Price each placement, then aggregate

For service costs, each placement is priced with its own state share, write share, replication factor, and region count. Network services are summed by type and retain a `placements` breakdown, so a region-confined keyspace contributes zero inter-region replication cost.

Backup snapshot bytes are computed per placement because replication factors may differ, then summed with `math.fsum` and rounded once for the plan. Daily upload volume is also summed per placement, while the retention/floor policy is applied once. This avoids both mixed-RF mispricing and per-placement rounding or minimum-charge inflation.

Candidate plans pass the original placement arguments through to service costing, so the costs used to rank a recommendation match the costs returned on that recommendation. Capacity sizing currently requires all supplied placements to share one replication factor; mixed factors are supported for cost attribution but deliberately rejected for sizing.

### Preserve the simple path

The existing `copies_per_region` API remains the simple whole-plan input. Explicit RF=1 is accepted as measured reality, producing no intra- or inter-region replication traffic. Backup eligibility is checked before backup arithmetic, preserving disabled-backup behavior even for otherwise unusable region shapes.

## Are there any tests?

Yes. The focused Cassandra suite passes with **89 tests**, covering placement derivation, partial write attribution, aggregate share validation, confined-keyspace traffic, candidate-plan propagation, RF constraints, mixed-RF backup snapshots, aggregate-before-rounding behavior, and compatibility of the single-replication path.

The full pre-commit pipeline passes: Ruff, formatting, Flake8, mypy, Pylint **10.00/10**, repository guards, and cost-baseline capture. The downstream ag-cass integration suite passes **107 tests** with this SCM branch installed editable. The final diff also received multi-model independent review with no blocking findings.

```python
placements = [
    KeyspacePlacement(
        keyspaces=["events"],
        copies_per_region=3,
        regions=["us-east-1", "us-west-2"],
        state_size_weight=1200,
        write_weight=40,
    ),
    KeyspacePlacement(
        keyspaces=["events_us_east_1"],
        copies_per_region=3,
        regions=["us-east-1"],
        state_size_weight=600,
        write_weight=90,
    ),
]

replication_by_region = keyspace_replication_by_region(placements)
extra_model_arguments = {
    "keyspace_replication": replication_by_region["us-east-1"],
}
```

Use the corresponding derived list for each regional plan. Callers must still verify that the returned region keys cover every deployed region; an omitted region would otherwise fall back to simple whole-plan pricing.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/cassandra.py
+++ b/service_capacity_modeling/models/org/netflix/cassandra.py
@@ -14,7 +14,6 @@ from typing import Union
 
 from pydantic import BaseModel
 from pydantic import Field
-from pydantic import field_validator
 from pydantic import model_validator
 
 from service_capacity_modeling.enum_utils import enum_docstrings
@@ -179,11 +178,6 @@ _CASSANDRA_SERIALIZATION_AMPLIFICATION = 1.42
 # Configurable via backup_retention_days in extra_model_arguments for clusters
 # with different compaction strategies (TWCS) or aggressive TTLs.
 _DEFAULT_BACKUP_RETENTION_DAYS = 14.0
-
-# Float slack allowed when checking that keyspace replication shares partition
-# a plan. Shares are computed by a caller dividing measured telemetry, so exact
-# equality to 1.0 is not achievable.
-_KEYSPACE_SHARE_TOLERANCE = 1e-6
 
 # Known Cassandra write size defaults (from default_desires() and interface.py).
 # Used to detect whether estimated_mean_write_size_bytes was user-supplied or
@@ -1282,6 +1276,11 @@ class KeyspaceReplication(BaseModel):
     with no cross-region replication. Since network cost scales with the regions
     a write must cross, pricing the whole cluster at one placement charges
     phantom cross-region replication to the single-region keyspaces.
+
+    Weights are relative, not fractions: pass measured magnitudes and the model
+    normalises them against each other. The plan's own totals stay the source of
+    truth, so a breakdown cannot contradict them no matter what scale the caller
+    works in, and there is no sum-to-one rule to get wrong.
     """
 
     keyspaces: List[str] = Field(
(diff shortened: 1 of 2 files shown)


homatthew:
## Decision: fix it in the model. No caller-side bridge.

Four independent model reviews were asked to choose between teaching the capacity model
about replication placements, versus doing the decomposition in the calling service with N
planner calls and no library change. **All four picked the model.**

### Why, in one fact

`service_costs(service_type, context, desires, extra_model_arguments)` takes **no instance,
no drive, no cluster**. And `plan_rank = compute_cost * (1 + total_penalty) + service_cost`.
Service cost is therefore the same additive constant on every candidate shape, and adding a
constant cannot move an argmin. Regret strips it too.

So a mis-attributed cross-region charge is a **reporting** error, not a **provisioning**
one. Nothing is mis-shaped by it, and nothing becomes mis-shaped by fixing it properly
rather than quickly. That removes the argument for a caller-side bridge.

### Why not the caller-side version — demonstrated, not argued

I built it as an offline script and it inflated a cluster's total service cost by ~80%,
with `net.intra.region` and `backup.*` each landing 4x high in a four-region cluster.

That is not a finding, it's the bug. Regional plans return an *additive share* so summing
all regions recovers the fleetwide cost, and `num_regions` is the divisor. Pass
`num_regions=1` for a single-region placement and that division becomes a no-op — that call
returns a fleet total while the whole-cluster call returns a per-region share. Sum them and
the mismatch is invisible: no exception, no failed assertion, just a plausible wrong number.

Encoding that normalization convention across a service boundary that doesn't document it,
doesn't version it, and has no test on either side is precisely the silent-wrong-number
class this change exists to remove.

Two further problems with the caller-side version: `extract_baseline` returns a whole
`CapacityPlan` whose `total_annual_cost` is a computed field, so a caller either publishes a
corrected total beside a plan that disagrees with it or mutates the planner's response; and
the fan-out grows from one call per region to one per region per placement.

### What changed in this PR as a result

**`state_size_share` is gone.** The backup snapshot is identical for `num_regions` of 1, 2
and 4 — placement does not affect it — and network cost is completely independent of state
size. Splitting state across placements was a no-op except for `math.ceil` rounding,
drifting the summed snapshot by ±1 GiB against the undivided plan.

Write rate is the only placement-sensitive input, so the model now splits exactly one number
and says so. That removes a validated field that could only ever introduce error, and halves
the share validator.

**The per-region scoping contract is now documented and tested**, because it is the thing I
got wrong above: a single-region entry is priced whole, not as a regional share, and a
caller must list it in one region's input only.

PR #315, opened 2026-08-07
Regenerate stale m7a/c7a/r7a/m6a cpu_ipc_scale 

hsane-dev:
# Summary
`deduce_cpu_ipc_scale()` was already fixed in this [PR](https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/283) to correctly compute `cpu_ipc_scale = arch_ipc * ht_factor`, but the shipped shape JSON for m7a/c7a/r7a predates that fix, so it never got the corrected value. Auditing this also surfaced m6a with the same staleness. This PR is a data regen + a regression test — no production logic changes.

## Changes
- **New test** `test_cpu_ipc_scale_invariant.py` asserts every `auto_*.json`'s `cpu_ipc_scale` matches `INSTANCE_TYPES[family]["cpu_ipc_scale"] * ht_factor`. Failed on exactly m7a/c7a/r7a/m6a before the regen, passes on all 41 auto-generated families after.
- **Regenerated** `auto_{m7a,c7a,r7a,m6a}.json` via `tools/auto_shape.py` . Only `cpu_ipc_scale` changed:
  - m7a/c7a/r7a: `1.5 → 1.95` (GENOA_IPC × 1.5 no-HT factor)
  - m6a: `1.0 → 1.15` (MILAN_IPC, HT-enabled so factor is 1.0)
- **Updated** `tests/test_common.py`: 3 `normalize_cores()` assertions that baked in m7a's old  IPC.
- **Regenerated** `baseline_costs.json` / `baseline_uncertain.json` via `capture_baseline_costs` (documented `tox -e capture-baseline` path) to reflect m7a's corrected cost/selection behavior.

New table should look like:
| Family              | Arch / HT     | Before (stale)     | After (corrected) | Formula                          |
|---------------------|---------------|--------------------|--------------------|-----------------------------------|
| m6a                 | Milan, HT-on  | 1.0                | 1.15               | `MILAN_IPC(1.15) x 1.0` (HT-on)    |
| m7a / c7a / r7a     | Genoa, no-HT  | 1.5                | 1.95               | `GENOA_IPC(1.2995) x 1.5` (HT-off) |
| m8a / c8a / r8a     | Turin, no-HT  | already correct    | 1.98               | `TURIN_IPC(1.325) x 1.5` (HT-off, unchanged, shown for reference) |

## Test plan
- Invariant test passes for all 41 families
- Full suite: 853 passed, 1 skipped, 0 failures

## Follow-up
This PR will be a part of a series of  PRs addressing the HT/IPC consistency cleanup


homatthew:
Is the implication that all m7a loads have been overprovisioning? Do you have worries about how this would affect the stateful fleet using EBS (e.g. cassandra)?

tests/tools/test_cpu_ipc_scale_invariant.py
```diff
@@ -0,0 +1,57 @@
+"""Ties shipped auto_*.json cpu_ipc_scale values back to INSTANCE_TYPES.
```
jasonk000, line 1: Good idea.

jasonk000 (approved):
Approved since it's definitely moving in the right direction - but -

I still find this a bit confusing to reason about. HT should make per-vcpu IPC slower, rather than making non-SMT faster. This semantic coupling really needs clarification.

hsane-dev:
> Is the implication that all m7a loads have been overprovisioning? Do you have worries about how this would affect the stateful fleet using EBS (e.g. cassandra)?

Yes — 7a was being scored weaker, so CPU-bound workloads could pick an oversized/pricier family. For EBS-bound like Cassandra specifically which is more disk bound, I dont believe recommendation should be affected or change.

homatthew:
`test_cpu_scale_up_disk_scale_down` used to pick `c6id.8xlarge` with local NVMe. With the `c7a` IPC multiplier changed from `1.5` to `1.95`, it picks `c7a.8xlarge` with EBS.

This makes sense given the relative improved performance of c6id -> c7a with this PR

homatthew (approved):

PR #316, opened 2026-08-08
Preserve valid deployed Cassandra EBS topology evidence

homatthew:
## Summary

The Cassandra model can overestimate EBS IOPS for a topology that is already running successfully. This change uses an aligned production observation to calibrate modeled IOPS, then applies the existing safety policy separately: target 90% disk-IOPS utilization and round the shared gp3 IOPS total to 200.

A comfortable observation may tune an overly conservative estimate downward. An observation at the configured IOPS limit is censored, so it can increase demand but can never tune the model downward.

## Input contract

Consumers provide one observation from the planning region:

```yaml
ebs_iops_evidence:
  peak_iops_per_node: 6000
  configured_iops_per_node: 16000
  observed_regional_workload:
    read_per_second: 10000
    write_per_second: 50000
    mean_read_size_bytes: 4096   # optional
    mean_write_size_bytes: 1024  # optional
```

`observed_regional_workload` is the current Cassandra traffic measured during the same window as `peak_iops_per_node`. It is deliberately separate from `query_pattern`, which is the target workload being planned. They are normally equal for status-quo planning but differ when planning for a forecast.

The workload must include some read or write traffic. Mean request sizes are optional. When absent, the planner uses the effective planning query-pattern sizes and reports that fallback in the explanation output.

The aligned workload stays bundled with the EBS observation in this first integration: splitting them would allow an IOPS peak to be paired with traffic from a different window, and composed systems can contain multiple model tiers with different effective workloads. This model-specific location is not a claim that observed workload is Cassandra-only. If another model needs the same kind of aligned runtime evidence, we should promote an optional generalized form into the shared current-deployment interface rather than introduce parallel model-specific representations.

Existing planner inputs remain authoritative for planning region, current instance and drive, node count, data per node, and zone layout. Configured IOPS remains explicit because it tells the planner whether the peak was capped and therefore unsafe to use for downward calibration.

## Approach

For a homogeneous current EBS deployment, the planner:

1. Models today's per-node IOPS from the observed regional workload and deployed topology.
2. Computes `calibration factor = observed current IOPS / modeled current IOPS`.
3. Applies that factor to each candidate's independently modeled IOPS demand.
4. Clamps the factor to at least 1 when the observation reached configured IOPS.
5. Applies the composed `disk_iops` buffer, then rounds the shared gp3 total once.

If current topology evidence is inconsistent, calibration is disabled and sizing is identical to planning without evidence. A matching candidate topology retains today's node-count floor. Other candidate shapes recompute node count and data density normally. Existing attached-drive sizing adds nodes when buffered demand exceeds gp3's maximum.

## Expected behavior

For the composed TimeSeries example, five `r7a.xlarge` nodes per zone observe 6,000 IOPS per node. The six-node candidate models 5,614.17 IOPS per node before calibration. The deployed observation maps this to 5,001.48 expected IOPS per node. At 90% target utilization, the plan requires 5,557.21 IOPS and provisions 5,600.

The explanation reports modeled current and candidate IOPS, raw and applied calibration factors, calibrated expected peak, buffer requirement, provisioned IOPS, resulting utilization, resolved observation sizes, and whether each size came from the observation or planning query pattern.

## Verification

- 12 focused Cassandra EBS evidence and TimeSeries composition tests passed after the final interface rename.
- Coverage includes comfortable downward calibration, censored-observation clamping, topology mismatch, family and drive filtering, composed buffers, total rounding, omitted-size fallback, and the five-to-six-node TimeSeries result.
- Full pre-commit passed: Ruff, formatting, YAML/JSON checks, Flake8, mypy, Pylint 10.00/10, enum guard, secret detection, and cost-baseline capture.


service_capacity_modeling/models/common.py
```diff
@@ -819,25 +923,50 @@ def sizing_for_count(node_count: int) -> AttachedDriveSizing:
...
-        candidate_sizing = sizing_for_count(candidate_count)
-        if candidate_sizing.next_count <= candidate_count:
-            final_count = candidate_count
-            final_sizing = candidate_sizing
-            break
-        final_count_source = (
-            "disk_iops"
-            if candidate_sizing.count_disk_iops >= candidate_sizing.count_disk_capacity
-            else "disk_capacity"
+    if max_attached_drive_io_per_s is None:
```
homatthew: Only the explicit-limit path guarantees the monotonic behavior needed for binary search. Existing unrestricted callers keep the exhaustive search because their callbacks may be non-monotonic.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -1453,6 +1969,110 @@ class NflxCassandraArguments(BaseModel):
...
+        description="Zones per region represented by the aligned EBS evidence.",
+    )
+    ebs_planning_baseline_num_regions: Optional[int] = Field(
+        default=None,
+        gt=0,
+        description="Regions represented by the aligned EBS evidence.",
+    )
+
+    @model_validator(mode="after")
+    def _check_ebs_topology_evidence(self) -> "NflxCassandraArguments":
```
homatthew: These fields are one evidence snapshot. Requiring the complete bundle prevents us from combining an IOPS peak with demand or topology from a different window.

PR #317, opened 2026-08-10
Add with_effective_cpu() helper to resize instances without corrupting HT signal

hsane-dev:
# Summary
Adds a helper, with_effective_cpu(), for safely shrinking an instance's vCPU count.

Instances track both cpu (vCPU count) and cpu_cores (real physical cores) — that's how we know if an instance is hyperthreaded. If code changes cpu alone without scaling cpu_cores to match, the instance ends up in a nonsense state that flips the hyperthreaded signal cpu_headroom_target() uses, silently producing the wrong capacity headroom.

This helper resizes both fields together, preserving the original ratio, so callers that need a smaller effective CPU count (e.g. penalizing a candidate instance) can't hit this bug.

# Testing
- New test passes; full existing test suite passes with no changes

# Notes
- This PR is independent of [#315](https://github.com/Netflix-Skunkworks/service-capacity-modeling/pull/315) — different files, no shared dependency, can merge in either order.
- This is groundwork only. It doesn't fix any live bug by itself; a follow-up PR in the instance-recommendation service will switch its wide-fanout logic to use this helper instead of its current manual copy.

### DIFF
--- a/service_capacity_modeling/models/common.py
+++ b/service_capacity_modeling/models/common.py
@@ -285,6 +285,21 @@ def _reserved_headroom(
     return strategy.calculate_reserved_headroom(effective_cpu)
 
 
+def with_effective_cpu(instance: Instance, effective_cpu: int) -> Instance:
+    """Return a copy of instance with its vCPU count adjusted, preserving HT ratio.
+
+    Callers that need to reason about a subset of an instance's vCPUs (for
+    example, penalizing a wide fan-out candidate down to a smaller effective
+    core count) must not change `cpu` without also scaling `cpu_cores` by the
+    same ratio: doing so silently flips the `cores < cpu` hyper-threading
+    signal that `cpu_headroom_target` relies on to pick the right headroom
+    multiplier.
+    """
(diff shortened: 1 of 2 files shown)


service_capacity_modeling/models/common.py
```diff
@@ -285,6 +285,21 @@ def _reserved_headroom(
     return strategy.calculate_reserved_headroom(effective_cpu)
 
 
+def with_effective_cpu(instance: Instance, effective_cpu: int) -> Instance:
```
jasonk000, line 288: I suspect relying on this here only benefits models that use it. I see the instance-recs PR, but, existing SCM models will not know about this?

jasonk000 (approved):
Yes. Should usage be expanded, or is this the only place it's used.

service_capacity_modeling/models/common.py
hsane-dev: no existing SCM model does this cpu-only resize today. Right now this is only used by the follow-up instancerecs [PR](https://github.netflix.net/corp/cie-instance-recommendation-service/pull/253) for widefanout.

PR #318, opened 2026-08-11
fix(read-only-kv): model SST memory overhead

william-schor:
## Summary

- model resident RocksDB SST table-reader metadata as an opt-in memory placement constraint
- retain legacy disk-only placement unless the workload supplies the named `read_only_kv_sst_memory` desired buffer
- enforce disk and memory as independent partition-placement constraints for opted-in workloads
- evaluate exact partition-density breakpoints and select the plan with the fewest nodes
- preserve higher replication factor as the tie-breaker for plans with equal node count

The named buffer must target `memory` or generic `storage`. Its ratio composes with other applicable memory and storage buffers, so a ratio of 1.2 applies 20% headroom to the configurable 5% SST estimate. Setting `sst_memory_overhead_ratio` to 0 remains an explicit disk-only kill switch.

Workloads without the named buffer do not publish an SST memory requirement, apply memory reservations, or change placement. Generated fleet baselines therefore remain identical to the existing disk-only baselines.

## Placement safety

For opted-in workloads, the planner:

- reserves the greater of the model minimum and workload app plus system reservations
- derives usable node memory with the existing memory abstraction
- applies explicit memory and generic storage buffers
- rejects shapes with non-positive usable memory
- reports buffered regional SST memory across all replicas
- exposes independent disk and memory partition ceilings
- preserves exact floating-point fits and skips equivalent partition densities

## Regression coverage

An anonymized large-manifest regression proves that the named buffer prevents one complete copy from being placed on a node when buffered SST metadata exceeds usable memory. Separate regressions prove that unnamed memory buffers do not activate the feature, the named buffer survives planner default merging, invalid components are rejected, the zero-ratio kill switch restores disk-only behavior, and lower-cost replica/partition combinations win.

## Verification

- Python 3.12 suite: 833 passed, 1 skipped
- focused partition-aware and read-only KV suite: 62 passed
- cost regression suite: 25 passed
- baseline capture: 19/19 certain and 3/3 uncertain scenarios
- generated cost baselines match the existing disk-only values
- Ruff lint and formatting: passed
- Python 3.12 mypy: passed
- targeted Pylint: 10/10

The aggregate local pre-commit command remains blocked by its existing Python 3.9/3.10 hooks parsing Python 3.12 type-alias syntax. The Python 3.12 checks above pass.

### DIFF
--- a/service_capacity_modeling/tools/data/baseline_costs.json
+++ b/service_capacity_modeling/tools/data/baseline_costs.json
@@ -1261,36 +1261,41 @@
   },
   {
     "annual_costs": {
-      "read-only-kv.regional-clusters": 434718.04
+      "read-only-kv.regional-clusters": 480930.32
     },
     "clusters": [
       {
-        "annual_cost": 434718.04,
+        "annual_cost": 480930.32,
         "cluster_params": {
-          "effective_disk_per_node_gib": 1900,
-          "read-only-kv.available_sst_memory_per_node_gib": 53.39,
+          "effective_disk_per_node_gib": 2355.2,
+          "read-only-kv.available_sst_memory_per_node_gib": 114.86,
+          "read-only-kv.buffered_sst_memory_per_copy_gib": 2880.0,
+          "read-only-kv.buffered_sst_memory_per_partition_gib": 5.625,
+          "read-only-kv.estimated_sst_memory_per_copy_gib": 2400.0,
+          "read-only-kv.max_partitions_per_node_by_disk": 21,
+          "read-only-kv.max_partitions_per_node_by_memory": 20,
+          "read-only-kv.memory_buffer_ratio": 1.2,
           "read-only-kv.min_replica_count": 4,
-          "read-only-kv.nodes_for_cpu": 10,
-          "read-only-kv.nodes_for_one_copy": 47,
-          "read-only-kv.partitions_per_node": 11,
+          "read-only-kv.nodes_for_cpu": 5,
+          "read-only-kv.nodes_for_one_copy": 26,
+          "read-only-kv.partitions_per_node": 20,
           "read-only-kv.replica_count": 4,
+          "read-only-kv.reserved_memory_per_node_gib": 8.0,
           "read-only-kv.sst_memory_overhead_ratio": 0.05,
-          "read-only-kv.sst_memory_per_copy_gib": 2400.0,
-          "read-only-kv.sst_memory_per_partition_gib": 4.6875,
-          "read-only-kv.total_num_partitions": 512,
-          "read-only-kv.total_sst_memory_gib": 9600.0
+          "read-only-kv.total_buffered_sst_memory_gib": 11520.0,
+          "read-only-kv.total_num_partitions": 512
         },
         "cluster_type": "read-only-kv",
-        "count": 188,
+        "count": 104,
         "deployment": "regional",
-        "instance": "i3.2xlarge"
+        "instance": "i3.4xlarge"
       }
     ],
     "model": "org.netflix.read-only-kv",
     "region": "us-east-1",
     "scenario": "read_only_kv_large",
     "service_tier": 1,
-    "total_annual_cost": 434718.04
+    "total_annual_cost": 480930.32
   },
   {
     "annual_costs": {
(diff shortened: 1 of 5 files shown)


service_capacity_modeling/tools/data/baseline_costs.json
```diff
@@ -1261,31 +1261,34 @@
   },
   {
     "annual_costs": {
-      "read-only-kv.regional-clusters": 149967.0
+      "read-only-kv.regional-clusters": 434718.04
     },
     "clusters": [
       {
-        "annual_cost": 149967.0,
+        "annual_cost": 434718.04,
```
homatthew: @william-schor were you aware of how big of a swing this would cause?

homatthew: for some cases I supppose

william-schor: addressed via optimizing through checking all possible PPN/RF pairs

homatthew (changes requested):
Automated adversarial review: **requesting changes**.

The direction is right: model resident SST memory as an independent placement constraint and keep effective disk disk-only. However, this version does not yet satisfy the key safety invariant: every accepted plan must fit modeled SST memory after the workload's reservations and requested headroom.

1. **The memory fit ignores existing reservation and buffer abstractions, so it can recommend an OOM-prone node.** `read_only_kv.py:231-234` subtracts only `args.reserved_memory_gib`. It ignores `DataShape.reserved_instance_app_mem_gib`, `reserved_instance_system_mem_gib`, and explicit `BufferComponent.memory`/`storage` buffers. I reproduced the same 26-node, 5-partitions/node plan with app reservations of 0, 40, and 120 GiB on a ~61 GiB instance. A requested 2x memory buffer also produced the same plan. The 120 GiB case should reject the shape outright.

   Please use the existing abstractions: reserve `max(args.reserved_memory_gib, app_reserve + system_reserve)`, derive usable memory with `usable_memory_gib`, reject non-positive usable memory, and apply `buffer_for_components(..., [BufferComponent.memory])` to the per-partition SST demand. The generic `storage` buffer already falls back to both disk and memory.

2. **The default plan has almost no estimation-error margin.** In the large baseline, each selected `i3.2xlarge` has 53.39 GiB modeled usable for SST readers and packs `11 * 4.6875 = 51.5625 GiB`: 96.58% utilization, leaving 1.8275 GiB. If the real coefficient is only 5.177% instead of 5%, or another unmodeled consumer uses >1.83 GiB, the node can OOM. The 5% coefficient must either be demonstrated to be a conservative upper bound including variance, or paired with an explicit, documented memory-headroom policy. Arithmetic fit against an uncalibrated point estimate is not an OOM guarantee.

3. **`CapacityRequirement.mem_gib` violates the regional-total contract.** The model exports one-copy SST memory, while `plan_comparison.py` compares that value with total regional provisioned memory. For the RF=2 regression, the requirement is 650 GiB while `read-only-kv.total_sst_memory_gib` is 1,300 GiB. That can falsely report an existing cluster as sufficient. Once RF is selected, export `mem_gib * replica_count` (without mutating the requirement inside the cluster-construction helper) and add a plan-comparison regression.

4. **The pushed search is exact in result but not computationally optimal.** Capping the loop at `min(max_ppn, n_partitions)` improves the previous version, but `total_num_partitions` has no upper bound and the loop constructs a Pydantic result for every valid density. A 1,000,000-partition case took 0.876 s locally. The exact quotient-breakpoint search returned the identical optimum in 0.000959 s:

   ```python
   ppn = max_ppn
   while ppn > 0:
       nodes_for_one_copy = math.ceil(n_partitions / ppn)
       # evaluate this candidate
       ppn = (n_partitions - 1) // nodes_for_one_copy
   ```

   This is exact for the current `(fewest nodes, highest RF, densest packing)` rank: every skipped density has the same `nodes_for_one_copy`, RF, and total nodes, while the evaluated breakpoint has the highest PPN in that block. Construct only the winning result after the search.

5. **Exact fits can be rejected by floating-point rounding.** With memory capacity `53.39` and per-partition memory `53.39 / 47`, the ratio evaluates to `46.99999999999999`; `int(...)` admits only 46 partitions and the current search returns no plan under an 8-node cap, although 47 is an exact fit. Normalize the capacity ratio by one ULP before flooring and add this regression.

Please also extend the property strategy to generate paired memory constraints; the current Hypothesis oracle exercises disk-only problems and would not catch deletion of the memory clamp.

The 2.90x baseline cost increase is explained by the new constraint, not by effective-disk inflation: the old `i3en.xlarge` plan packed 21 partitions/node and used 100 nodes at $149,967/year; the new memory-limited `i3.2xlarge` plan packs 11, needs `ceil(512/11) * RF4 = 188` nodes, and costs $434,718/year. That may be the correct cost if 5% plus calibrated headroom is the accepted requirement, but the current implementation does not yet establish that safety premise.

I recommend the independent disk-fit/memory-fit design with the exact breakpoint search, generic reservations and buffers, total-memory requirement semantics, and explicit memory headroom. With those changes and regressions, I would approve this direction.

homatthew:
Automated follow-up review: I put the proposed safety repair in #319 so this branch stays untouched and you can choose whether to merge it.

#319 keeps your min-node optimizer and independent disk/memory fit, then adds the existing app/system reservation and usable-memory abstractions, memory and generic storage buffers, a default 1.2 SST-memory buffer, exact quotient-breakpoint search, exact-float-fit handling, RF-scaled regional memory requirements, explicit per-resource partition ceilings, and disk-only rollback when the SST ratio is zero.

The large baseline becomes 104 i3.4xlarge nodes at ,930.32/year. That is 10.6% above this PR's current ,718.04 plan and 3.21x the old disk-only ,967 baseline. The selected nodes use 112.5 GiB of buffered SST allocation out of 114.86 GiB available, so the raw 5% estimate can rise to about 6.126% before exhausting modeled usable memory.

Verdict on the current #318 head: NEEDS CHANGES for OOM safety. Verdict with #319 merged: APPROVE, subject to #319 CI. The 5% coefficient is still a planning assumption that should be calibrated separately against production RocksDB resident memory.

homatthew:
Automated follow-up: #319 is now green on Python 3.10, 3.11, and 3.12 and has an AI APPROVE verdict. The conditional approval in my prior comment is satisfied if #319 is merged into this branch.

william-schor:
@homatthew #319 is merged into this branch. I added a narrow rollout gate on top: SST memory placement now activates only when the workload supplies the named `read_only_kv_sst_memory` desired memory buffer. Workloads without that buffer retain the existing disk-only plans and baseline costs. Opted-in workloads keep #319's reservations, composed memory/storage headroom, regional memory contract, and hardened optimizer.

The current commit is locally green on the full Python 3.12 suite, focused regressions, baseline capture, cost regressions, Ruff, formatting, mypy, and targeted Pylint. Would you re-review the combined behavior when CI completes?

PR #319, opened 2026-08-12
fix(read-only-kv): harden SST memory placement

homatthew:
## Summary

This is an opt-in safety patch for #318. It leaves Wschor's branch unchanged and can be merged into that branch if the additional OOM protection and cost are acceptable.

The patch keeps disk and SST memory as independent placement constraints. It applies the existing memory abstractions to the 5% SST table-reader estimate:

- `buffer_for_components(..., [BufferComponent.memory])`, including generic `storage` fallbacks
- `max(reserved_memory_gib, app reservation + system reservation)`
- `usable_memory_gib`, with non-positive usable memory rejected
- a default 1.2 memory buffer, so the 5% estimate consumes at most 83.3% of its modeled allocation

`sst_memory_overhead_ratio=0` still gives disk-only placement and bypasses SST-specific memory rejection.

## Placement algorithm

The exact search evaluates one candidate for each distinct `nodes_for_one_copy` quotient:

```python
ppn = (n_partitions - 1) // nodes_for_one_copy
```

Every skipped density has the same nodes per copy, replica count, and node count. The evaluated density wins the existing dense-packing tie-break, so the result is identical to exhaustive search while running in O(sqrt(partitions)) candidates. A one-ULP normalization preserves exact floating-point fits.

The planner reports the independent disk and memory partition ceilings rather than a single ambiguous `capacity_limited_by` label. Raw estimated SST memory and buffered SST allocation also have distinct parameter names.

## Memory contract

The placement calculation uses buffered memory for one complete copy. After the algorithm selects RF, `CapacityRequirement.mem_gib` reports the buffered regional total across all copies. This matches the regional requirement contract used by plan comparison.

The 5% coefficient remains a configurable planning assumption, not a measured upper bound. The default 1.2 buffer protects the modeled fit against a coefficient increase from 5% to about 6.126% in the large baseline, a 22.5% margin. Calibration against production RocksDB resident memory remains a separate decision.

## Cost impact

| Version | Large baseline | Annual cost |
| --- | --- | ---: |
| Existing disk-only baseline | 100 × `i3en.xlarge` | $149,967.00 |
| #318 current head | 188 × `i3.2xlarge` | $434,718.04 |
| This safety patch | 104 × `i3.4xlarge` | $480,930.32 |

This patch adds 10.6% over #318. The roughly 3.2x increase versus the existing baseline comes from admitting SST memory as a real bin-packing constraint; the incremental increase here is the explicit 20% memory buffer. Effective disk is unchanged by the SST ratio.

For the selected large plan, disk allows 21 partitions per node and buffered SST memory allows 20. Each node has 114.86 GiB available for SST readers and receives 112.5 GiB of buffered SST allocation.

## Verification

- Full Python 3.12 suite: 829 passed, 1 skipped
- Focused partition-aware and read-only KV suite: 60 passed
- `tox -e pre-commit`: passed, including Ruff, mypy, pylint 10/10, and baseline capture
- Baseline capture: 19/19 certain and 3/3 uncertain scenarios
- Partition algorithm coverage: 100%; read-only KV coverage: 97% focused, 98% full suite
- Independent exhaustive-oracle reviews found no optimizer mismatch across randomized disk-only and memory-constrained problems
- Independent post-fix review: approved with no remaining actionable findings

### DIFF
--- a/service_capacity_modeling/models/org/netflix/partition_aware_algorithm.py
+++ b/service_capacity_modeling/models/org/netflix/partition_aware_algorithm.py
@@ -49,6 +49,14 @@ class CapacityResult(BaseModel):
     replica_count: int  # Replication factor
     partitions_per_node: int  # Partitions placed on each node
     nodes_for_one_copy: int  # Nodes needed for one complete copy of data
+    max_partitions_per_node_by_disk: int
+    max_partitions_per_node_by_memory: Optional[int] = None
+
+
+def _partitions_that_fit(capacity_gib: float, partition_size_gib: float) -> int:
+    """Floor a capacity ratio without losing an exact fit to float error."""
+    fit = capacity_gib / partition_size_gib
+    return math.floor(math.nextafter(fit, math.inf))
 
 
 def search_for_min_nodes(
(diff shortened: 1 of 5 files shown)


homatthew (commented):
Automated AI review verdict: APPROVE.

No actionable findings remain. The optimizer matched exhaustive randomized oracles; modeled SST memory respects existing reservation and buffer abstractions; the feature-off path remains disk-only; raw and buffered memory contracts are explicit; full local verification passed; and GitHub CI passed on Python 3.10, 3.11, and 3.12.

The 5% SST coefficient remains a configurable planning assumption that should be calibrated separately. That caveat does not block this safety patch.

PR #321, opened 2026-08-19
feat(time-series): default the Cassandra tier to EBS

kelebra:
Move TimeSeries capacity planning to only provision EBS instancy type for vast majority of workloads to simplify support an maintenance.

### DIFF
--- a/service_capacity_modeling/models/org/netflix/time_series.py
+++ b/service_capacity_modeling/models/org/netflix/time_series.py
@@ -22,6 +22,42 @@ from service_capacity_modeling.interface import RegionContext
 from service_capacity_modeling.models import CapacityModel
 
 
+# These bailouts send the Cassandra tier back to regular local disk planning.
+CASSANDRA_EBS_MAX_READ_PER_SECOND = 200_000
+CASSANDRA_EBS_MIN_STATE_SIZE_GIB = 1024
+
+# Ordered bailout rules, each a (name, predicate over the Cassandra facing
+# desires) pair. The first rule that holds wins and names the reason we stayed
+# on local disks.
+CASSANDRA_EBS_BAILOUTS: Tuple[Tuple[str, Callable[[CapacityDesires], bool]], ...] = (
+    (
+        "read_throughput",
+        lambda desires: desires.query_pattern.estimated_read_per_second.mid
+        > CASSANDRA_EBS_MAX_READ_PER_SECOND,
+    ),
+    (
+        "state_size",
+        lambda desires: desires.data_shape.estimated_state_size_gib.mid
+        < CASSANDRA_EBS_MIN_STATE_SIZE_GIB,
+    ),
+)
+
+
+def cassandra_ebs_bailout(cassandra_desires: CapacityDesires) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


service_capacity_modeling/models/org/netflix/time_series.py
```diff
@@ -22,6 +22,42 @@
...
+
+    Pass the desires Cassandra will actually be planned with, so after
+    TimeSeries read amplification: the read rule is about load landing on
+    Cassandra, not load landing on the TimeSeries tier. An unstated dataset
+    size reads as zero and therefore bails out to local disks. Returns None
+    when nothing bails out and the tier should plan on EBS.
+    """
+    for name, applies in CASSANDRA_EBS_BAILOUTS:
+        if applies(cassandra_desires):
+            return name
```
rajivshringi: Do you intend to use this return value upstream? I don't see this logged anywhere

service_capacity_modeling/models/org/netflix/time_series.py
```diff
@@ -82,6 +118,12 @@ def _modify_elasticsearch_desires(
             )
             return relaxed
 
+        # require_local_disks defaults true and on its own excuses every EBS-only
+        # instance, so asking for EBS has to clear it too.
+        if cassandra_ebs_bailout(_modify_cassandra_desires(user_desires)) is None:
+            extra_model_arguments["require_local_disks"] = False
+            extra_model_arguments["require_attached_disks"] = True
```
rajivshringi, line 105: Do we need both overrides?

kelebra: `require_local_disks=False` permits EBS (it defaults `true` and excuses every EBS-only instance); `require_attached_disks=True` requires it, without both we are allowing local disks to participate.

service_capacity_modeling/models/org/netflix/time_series.py
kelebra: Inlined both conditions and dropped the table, the helper, and test_bailout_rules_are_extensible (it tested the table's extensibility, not planner behavior).

rajivshringi (approved):

homatthew:
I suggest keeping workload-derived state out of the shared extra_model_arguments object. The other composed models I checked, including KeyValue and GraphKV, express workload-derived child behavior through the CapacityDesires transform returned by compose_with. The dictionary mutations elsewhere are constant setdefault("root_disk_gib", 20) defaults, which are idempotent rather than dependent on one call’s desires.

These assignments are sticky across independent calls because the planner preserves a non-empty dictionary by identity. I reproduced an eligible TimeSeries call followed by a >200k-read call with the same dictionary: the second call stayed on gp3, while the same second call with a fresh dictionary selected local c5d disks.

Could Cassandra own a preference flag such as prefer_attached_disks instead? The existing require_local_disks and require_attached_disks fields are hard filters. A Cassandra-owned preference could evaluate the already-amplified Cassandra desires, rank EBS ahead when eligible, and retain local disks as a fallback. TimeSeries would only opt into that policy. This does not by itself solve argument ownership because compose_with has no per-child argument channel, so I would still isolate extra_model_arguments per public planner invocation and add a reused-dictionary regression test.

PR #323, opened 2026-08-23
Prefer EBS for Cassandra through maintenance regret

homatthew:
## Summary

PR #321 moved most TimeSeries Cassandra tiers to EBS using a caller-owned hard requirement. This change moves that operational preference into Cassandra so direct Cassandra, TimeSeries, and KeyValue planning share one policy without mutating caller arguments or excluding local NVMe.

Attached storage can grow and be replaced independently of compute. Replacing a local-disk node also moves its data, so the preference grows with data per node rather than total dataset size or fleet width.

## Approach

- Consider attached and local storage by default.
- Add a 20% compute-cost penalty to local-disk plans through 300 GiB per node.
- Increase it linearly to 30% at 1 TiB per node, then cap it.
- Apply the preference during deterministic ranking and uncertain planning.
- Compound planner family preferences with model-adjusted compute rank.
- Remove TimeSeries mutation of `extra_model_arguments`.

## Policy interface

`ephemeral_maintenance_regret` is the single caller control and defaults to `0.2`. Setting it to `0` disables the preference.

The 300 GiB baseline and 1 TiB cap are Cassandra policy, not caller tuning knobs. Existing `require_local_disks` and `require_attached_disks` arguments remain available when storage type is a hard requirement.

## Expected behavior

EBS wins when local compute savings do not clear the bounded maintenance penalty. Wide local fleets with little data per node receive only the baseline penalty. Local NVMe remains eligible and can still win when its compute advantage is large enough.

## Verification

- All 115 tests across the six touched Cassandra, TimeSeries, and KeyValue suites pass under Python 3.12.
- The tests cover curve boundaries, disabling the preference, per-node density, deterministic and uncertain planning, and composed-model behavior.
- Mypy and the complete pre-commit suite pass.
- Baseline capture matches the committed snapshots.

### DIFF
--- a/service_capacity_modeling/capacity_planner.py
+++ b/service_capacity_modeling/capacity_planner.py
@@ -91,7 +91,8 @@ class PlannerArguments(ExcludeUnsetModel):
         default=0.15,
         description="Rank inflation applied to plans using families outside "
         "model.preferred_families(). A non-preferred family must be this fraction "
-        "cheaper on compute cost to rank above a preferred alternative. "
+        "cheaper on penalty-adjusted compute cost to rank above a preferred "
+        "alternative. "
         "Set to 0.0 to evaluate all families by pure cost.",
     )
 
@@ -915,13 +916,11 @@ class CapacityPlanner:
                         service_cost = sum(
                             s.annual_cost for s in plan.candidate_clusters.services
                         )
-                        compute_cost = (
-                            plan.candidate_clusters.total_annual_cost - service_cost
-                        )
+                        ranked_compute_cost = plan.rank - service_cost
                         plan = plan.model_copy(
                             update={
                                 "rank": plan.rank
-                                + compute_cost * pargs.preferred_family_penalty
+                                + ranked_compute_cost * pargs.preferred_family_penalty
                             }
                         )
                     plans.append(plan)
--- a/service_capacity_modeling/models/org/netflix/cassandra.py
+++ b/service_capacity_modeling/models/org/netflix/cassandra.py
@@ -84,6 +84,8 @@ CRITICAL_TIERS: Set[int] = {0, 1}
 CRITICAL_TIER_MIN_CLUSTER_SIZE = 2
 EBS_HOTTER_BUFFER_RATIO = 0.5
 CASSANDRA_MAX_DISK_UTILIZATION = 0.55
+_EPHEMERAL_MAINTENANCE_THRESHOLD_GIB_PER_NODE = 300
+_EPHEMERAL_MAINTENANCE_CAP_GIB_PER_NODE = 1024
 
 
 def _with_disk_utilization_buffer(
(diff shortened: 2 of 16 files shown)


service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -1842,20 +1887,25 @@ def regret(
...
+                sum(c.annual_cost for c in proposed_plan.candidate_clusters.zonal)
+            )
+            for name in (
+                "large_instance",
+                "ephemeral_maintenance",
+            ):
+                if name not in penalties:
+                    continue
+                # Apply regret only to compute costs (same rationale as rank).
+                regrets[name] = penalties[name] * compute_cost
```
homatthew, line 1902: Uncertain planning does not use deterministic rank order, so apply the stored penalty here to preserve the same storage preference.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -1353,6 +1378,22 @@ class NflxCassandraArguments(BaseModel):
         "effective sort cost. Prevents AWS pricing rounding from favoring "
         "larger instances. Set to 0 to disable.",
     )
+    ephemeral_maintenance_regret: float = Field(
```
homatthew, line 1385: This preference lives in Cassandra so direct, TimeSeries, and KeyValue plans share it. Callers can still use `require_local_disks` or `require_attached_disks` when storage is a hard requirement.

service_capacity_modeling/capacity_planner.py
```diff
@@ -915,13 +916,11 @@ def _plan_certain(  # pylint: disable=too-many-positional-arguments
...
                         )
-                        compute_cost = (
-                            plan.candidate_clusters.total_annual_cost - service_cost
-                        )
+                        ranked_compute_cost = plan.rank - service_cost
                         plan = plan.model_copy(
                             update={
                                 "rank": plan.rank
-                                + compute_cost * pargs.preferred_family_penalty
+                                + ranked_compute_cost * pargs.preferred_family_penalty
```
homatthew, line 923: The family preference is another percentage preference, so apply it to the model-adjusted compute rank rather than adding it to raw compute. This keeps independently motivated penalties multiplicative and leaves service costs out because they do not vary with instance family.

service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -712,6 +716,27 @@ def _compute_penalties(
...
+    # Local disks must provide meaningful savings even at low density, and the
+    # required savings grow with data per node because replacement streams more
+    # data. The gradual ramp keeps local NVMe competitive until a node owns about
+    # 1 TiB, where the premium caps at 1.5 times the baseline.
+    if instance.drive is not None and ephemeral_maintenance_regret > 0:
+        density_fraction = min(
+            1,
+            max(
+                0,
+                (data_per_node_gib - _EPHEMERAL_MAINTENANCE_THRESHOLD_GIB_PER_NODE)
```
homatthew, line 729: The 300 GiB and 1 TiB points are fixed Cassandra policy here, not caller tuning knobs. `ephemeral_maintenance_regret` remains the single public control: callers may scale or disable the preference without redefining the density curve.

PR #326, opened 2026-08-30
Add Valkey ElastiCache capacity model

kelebra:
## Summary

- add `org.netflix.valkey` for managed Valkey clusters on ElastiCache
- support seventh- and eighth-generation managed node families
- size memory from explicit item count, explicit state, or WPS + TTL; read/write sizes represent total key-plus-value bytes
- use 16-byte keys and 50-byte values by default
- apply a benchmark-derived Valkey 8.1 per-item overhead curve, and expose 75% of each node's RAM as usable cache storage
- model ordinary and Lua operations as weighted consumers of the same single-thread CPU capacity
- use network bandwidth as both a primary-shard and total-node constraint
- require at least one read replica for durable clusters and enforce AWS's five-replicas-per-shard maximum
- enforce AWS's default 90-node cluster quota, with an explicit override for approved quota increases

## Why some topology choices look surprising

The planner evaluates the complete valid topology instead of assuming that taller nodes or read replicas are always the next scaling step:

- Taller nodes add memory and network bandwidth but do not increase per-shard operation throughput, so write-heavy workloads add shards even when a larger node appears to be the obvious choice.
- With one shard, a read replica can add read capacity with one node, while adding a durable shard requires both a primary and its replica.
- Reads and keys are assumed uniform across shards. Every shard therefore needs the same replica count; uneven replicas would leave the least-replicated shard overloaded.
- Once memory or writes already require several shards, adding another shard can require fewer total nodes than adding a replica to every existing shard.
- A shard can have at most five read replicas. Once that ceiling is reached, additional read demand must expand the shard count.
- The benchmarked per-item memory curve falls within a fitted bucket and jumps when a larger item crosses the next observed boundary. Values above the measured 128-byte curve continue through the fitted bucket progression and are flagged as extrapolated in plan context.
- Generation and node size are evaluated together: a memory- or network-dominated workload can favor a taller eighth-generation node, while a throughput-dominated workload can favor more small seventh-generation nodes.

## Input contract

| Input | Semantics |
|---|---|
| `query_pattern.estimated_read_per_second` | Ordinary reads per second |
| `query_pattern.estimated_write_per_second` | Total writes per second before applying the Lua share |
| `query_pattern.estimated_mean_read_size_bytes` | Total key-plus-value bytes transferred by one read; defaults to 66 bytes |
| `query_pattern.estimated_mean_write_size_bytes` | Total key-plus-value bytes stored by one write; defaults to 66 bytes |
| `data_shape.estimated_state_item_count` | Preferred state input when item count is known |
| `data_shape.estimated_state_size_gib` | State input when total uncompressed key-plus-value size is known |
| `data_shape.estimated_compression_ratio` | Value compression ratio; keys remain uncompressed; defaults to 1 |
| `data_shape.durability_slo_order` | Values at or above 1,000 require at least one replica per shard |
| `valkey.key_size_bytes` | Key portion of the total item size; defaults to 16 bytes |
| `valkey.ttl` | Positive finite ISO-8601 TTL used with WPS when neither item count nor state size is supplied; defaults to `PT24H` |
| `valkey.lua_write_percent` | Fraction of writes that execute Lua, from 0 to 1; defaults to 0 |
| `valkey.max_nodes_per_cluster` | Cluster node ceiling; defaults to 90 and can reflect an approved quota increase |

State sizing uses item count first, then explicit state size, then `WPS × TTL`.

### Small payload examples

Explicit item count with 16-byte keys and 37-byte values:

```python
desires = CapacityDesires(
    query_pattern=QueryPattern(
        estimated_read_per_second=certain_int(200_000),
        estimated_write_per_second=certain_int(100_000),
        estimated_mean_read_size_bytes=certain_int(53),
        estimated_mean_write_size_bytes=certain_int(53),
    ),
    data_shape=DataShape(
        estimated_state_item_count=certain_int(17_000_000),
        durability_slo_order=FixedInterval(low=10_000, mid=10_000, high=10_000),
    ),
)
extra_model_arguments = {
    "valkey.key_size_bytes": 16,
    "valkey.lua_write_percent": 1.0,
}
```

TTL-derived state using the default 16-byte key and 50-byte value:

```python
desires = CapacityDesires(
    query_pattern=QueryPattern(
        estimated_read_per_second=certain_int(200_000),
        estimated_write_per_second=certain_int(100_000),
    ),
)
extra_model_arguments = {
    "valkey.ttl": "PT1H",
    "valkey.lua_write_percent": 0.25,
}
```

## Weighted operation model

For a `cache.r7g` node, the model uses 700k ordinary operations/second and 50k Lua operations/second. Required CPU units are:

```text
ordinary reads / 700k
+ non-Lua writes / 700k
+ Lua writes / 50k
```

## Example least-regret recommendations

These recommendations use fixed inputs through the uncertain planner and report `least_regret[0]`.

| Durable | State / write workload | RPS | Instance | Topology | Binding behavior |
|---|---|---:|---|---|---|
| Yes | 1 GiB · 500k WPS · 0% Lua | 500k | `cache.r7g.large` | 1 shard · 1 replica / shard | Combined ordinary-operation load requires the second node |
| Yes | 1 GiB · 200k WPS · 25% Lua | 500k | `cache.r8g.large` | 1 shard · 1 replica / shard | Faster core keeps the weighted write load within one shard |
| Yes | 1 GiB · 100k WPS · 100% Lua | 1M | `cache.r7g.large` | 2 shards · 1 replica / shard | Two primaries are fully occupied by 50k Lua each; two replicas serve 500k reads each |
| Yes | 1 GiB · 300k WPS · 0% Lua | 5M | `cache.r7g.large` | 2 shards · 3 replicas / shard | Read demand exceeds one shard's replica ceiling |
| Yes | 300 GiB · 1k WPS · 0% Lua | 1k | `cache.r8g.2xlarge` | 11 shards · 1 replica / shard | Per-item overhead and 75% usable node RAM make memory the binding constraint |
| Yes | 1 GiB · 0 WPS · 0% Lua | 4,740,001 | `cache.r7g.large` | 2 shards · 3 replicas / shard | One shard cannot exceed five replicas |
| Yes | 1 GiB · 1.5M WPS · 0% Lua | 4.3M | `cache.r8g.large` | 2 shards · 3 replicas / shard | Faster core satisfies the weighted demand with two primary shards |

## Validation

- `tox -e pre-commit`
- focused Valkey model and universal property coverage
- explicit item-count preservation and positive finite TTL validation
- Valkey 8.1 benchmark-curve boundaries, including extrapolation reporting
- usable node memory, CPU and network sizing, durability, uniform sharding, and replica-limit expansion
- default cluster quota rejection and approved-quota override

### DIFF
--- a/service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_elasticache-on-demand.json
+++ b/service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_elasticache-on-demand.json
@@ -3,46 +3,46 @@
     "drives": {},
     "instances": {
       "cache.r7g.12xlarge": {
-        "annual_cost": 16509.1
+        "annual_cost": 36686.88
       },
       "cache.r7g.16xlarge": {
-        "annual_cost": 22015.28
+        "annual_cost": 48922.85
       },
       "cache.r7g.2xlarge": {
-        "annual_cost": 2753.09
+        "annual_cost": 6117.98
       },
       "cache.r7g.4xlarge": {
-        "annual_cost": 5503.03
+        "annual_cost": 12228.96
       },
       "cache.r7g.8xlarge": {
-        "annual_cost": 11009.22
+        "annual_cost": 24464.93
       },
       "cache.r7g.large": {
-        "annual_cost": 690.64
+        "annual_cost": 1534.75
       },
       "cache.r7g.xlarge": {
-        "annual_cost": 1378.12
+        "annual_cost": 3062.5
       },
       "cache.r8g.12xlarge": {
-        "annual_cost": 16608.43
+        "annual_cost": 36907.63
       },
       "cache.r8g.16xlarge": {
-        "annual_cost": 22144.97
+        "annual_cost": 49211.05
       },
       "cache.r8g.2xlarge": {
-        "annual_cost": 2768.07
+        "annual_cost": 6151.27
       },
       "cache.r8g.4xlarge": {
-        "annual_cost": 5536.14
+        "annual_cost": 12302.54
       },
       "cache.r8g.8xlarge": {
-        "annual_cost": 11072.29
+        "annual_cost": 24605.09
       },
       "cache.r8g.large": {
-        "annual_cost": 692.22
+        "annual_cost": 1538.26
       },
       "cache.r8g.xlarge": {
-        "annual_cost": 1384.04
+        "annual_cost": 3075.64
       }
     },
     "services": {},
--- a/service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_elasticache.json
+++ b/service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_elasticache.json
@@ -1,0 +1,51 @@
+{
+  "us-east-1": {
+    "drives": {},
+    "instances": {
+      "cache.r7g.12xlarge": {
+        "annual_cost": 16509.1
+      },
+      "cache.r7g.16xlarge": {
+        "annual_cost": 22015.28
+      },
+      "cache.r7g.2xlarge": {
+        "annual_cost": 2753.09
+      },
+      "cache.r7g.4xlarge": {
+        "annual_cost": 5503.03
+      },
+      "cache.r7g.8xlarge": {
+        "annual_cost": 11009.22
+      },
+      "cache.r7g.large": {
+        "annual_cost": 690.64
+      },
+      "cache.r7g.xlarge": {
+        "annual_cost": 1378.12
+      },
+      "cache.r8g.12xlarge": {
+        "annual_cost": 16608.43
+      },
+      "cache.r8g.16xlarge": {
+        "annual_cost": 22144.97
+      },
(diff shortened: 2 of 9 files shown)


service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_elasticache-on-demand.json
```diff
@@ -0,0 +1,51 @@
+{
+  "us-east-1": {
+    "drives": {},
+    "instances": {
```
kartik894, line 4: Any reason why we need two instance families here (r7g vs r8g)?

kartik894 (approved):
Great PR! One minor comment and 🚢

service_capacity_modeling/hardware/profiles/pricing/aws/3yr-reserved_elasticache-on-demand.json
kelebra: Synced offline, good to proceed for now with 2 proposed instance typed.

PR #328, opened 2026-09-03
Bound capacity planner CI fan-out

homatthew:
## Why

Cassandra now considers both local and attached storage by default. A representative uncertain plan still evaluates 5,856 hardware scenarios, but 3,328 attached-storage candidates now run the full estimator instead of exiting early. Tests that requested 128 to 1,024 simulations multiplied that candidate fan-out until Python CI jobs took 41 to 59 minutes.

The fix bounds each test at the contract it actually exercises and removes repeated work inside the Cassandra candidate loop. It does not raise Hypothesis deadlines or globally lower production planner defaults.

## Coverage changes

The universal property suite collected seven properties across 22 models. It now collects four, removing 66 pytest parametrizations, but only one behavioral contract was removed.

| Previous property | What changed | Coverage effect |
| --- | --- | --- |
| Positive cost | Folded into plan validity | Still 20 Hypothesis examples per model |
| Positive instance count | Folded into plan validity | Still 20 Hypothesis examples per model |
| Plan exists | Retained in the same test | No reduction |
| Cost increases with QPS | Retained unchanged | No reduction |
| Tier capacity relationship | Retained unchanged | No reduction |
| Determinism | Retained with two plans instead of three | Two observations are sufficient to assert equality |
| Installed CPU increases with QPS | Removed | Intentional coverage reduction; this is not a planner invariant |

Positive cost and instance count now inspect the plan already produced by the validity property instead of independently traversing the full catalog. Hypothesis still generates exactly 20 inputs per model for every retained assertion.

Installed CPU monotonicity was invalid at the winning-plan boundary. For one generated Cassandra case:

```text
Low QPS:   43.2 normalized cores, EBS, $6,424
High QPS:  42.0 normalized cores, local disk, $8,229
```

The higher workload selected a more efficient hardware and storage combination. Capacity and cost increased appropriately while installed CPU decreased slightly. The remaining risk is narrower: a model could accidentally calculate lower theoretical CPU demand for higher QPS while another constraint still makes the final plan more expensive. That should be covered as an opt-in model property by holding candidate family and storage type constant and comparing calculated `cpu_cores` requirements, not the winning plan's installed CPU. Cassandra already has targeted CPU scale-up/down coverage; EVCache, Java, Elasticsearch, and planner comparison paths also have model-specific CPU tests.

Other reductions preserve their actual contracts:

- Counter storage routing uses deterministic planning instead of 256 irrelevant uncertainty samples.
- Composition uses four samples instead of the 128-simulation default while still comparing Cassandra subplans and regrets byte-for-byte.
- Reproducibility performs two identical executions, the minimum needed to prove equality.
- Family diversity compares 12 with 64 samples instead of 12 with 1,024 and holds storage class constant.
- The deleted `conftest.py` was an ineffective fixture, not a test. Test modules had already imported their planner references before the fixture replaced the module attribute.

## Runtime changes

- Replace three whole-`CapacityDesires` deep copies per Cassandra hardware candidate with copies only at the nested mutation boundary.
- Make expensive uncertain-planner tests declare the simulations their assertion needs.
- Fold compatible assertions over the same generated plan rather than recalculating it.

Measured locally:

- Cassandra 12-simulation plan: 2.24s to 1.30s.
- Profiled calls: 38.1M to 7.0M.
- `deepcopy` cumulative time: 4.74s to 0.18s.
- Four pathological tests: about 50 aggregate CI minutes to 2.72s.
- Compositional identity under coverage: 61.99s to 3.62s.
- Full Python 3.12 tox job: 2m57s for 876 tests, 90% coverage, and mypy.

## Verification

- `tox -e py312`
- `tox run-parallel -e py310,py311`
- `tox -e pre-commit`
- GitHub Actions: Python 3.10 in 17m44s, 3.11 in 11m49s, and 3.12 in 16m59s: https://github.com/Netflix-Skunkworks/service-capacity-modeling/actions/runs/33785193393
- Independent read-only review found no surviving correctness or test-contract issue.


service_capacity_modeling/models/org/netflix/cassandra.py
```diff
@@ -112,24 +111,31 @@ def _with_disk_utilization_buffer(
         ratio=disk_utilization_ratio,
         components=[BufferComponent.disk],
     )
-    capped_desires.buffers.desired = {
-        **capped_desires.buffers.desired,
-        CASSANDRA_DISK_UTILIZATION_LIMIT: disk_utilization_buffer,
-    }
-    return capped_desires
+    buffers = desires.buffers.model_copy(
```
homatthew, line 114: This is the hot-path optimization. These transformations run once per viable hardware candidate, but only the nested buffer mapping changes. Copying that mutation path keeps candidates isolated while avoiding three full `CapacityDesires` deep copies per candidate. In the representative profile, cumulative `deepcopy` time fell from 4.74s to 0.18s.

tests/netflix/test_all_models_properties.py
```diff
@@ -104,113 +109,25 @@ def test_all_models_produce_valid_plans(model_name, data):
...
-    )
-
-    high_qps_desires = CapacityDesires(
-        service_tier=1,
-        query_pattern=QueryPattern(
-            estimated_read_per_second=certain_int(high_read_qps),
-            estimated_write_per_second=certain_int(high_write_qps),
-        ),
-        data_shape=DataShape(estimated_state_size_gib=certain_int(data_gib)),
+    cost = get_total_cost(plan)
```
homatthew, line 112: Positive cost and positive instance count were folded into this valid-plan property. This removes 66 collected pytest parametrizations, not 66 generated checks: all three assertions still inspect 20 Hypothesis examples per model, but now share one planner execution per example.

tests/netflix/test_all_models_properties.py
```diff
@@ -104,113 +109,25 @@ def test_all_models_produce_valid_plans(model_name, data):
...
-# ============================================================================
-# Universal Property: QPS Monotonicity
-# ============================================================================
-
-
-@settings(
-    max_examples=15, deadline=20000, suppress_health_check=[HealthCheck.filter_too_much]
-)
-@given(data=st.data())
-def test_all_models_scale_cpu_with_qps(model_name, data):
```
homatthew, line 117: Removing installed-CPU monotonicity is intentional. The winning plan may change instance family or storage type as QPS increases. One valid Cassandra example moved from 43.2 normalized cores on EBS at $6,424 to 42.0 cores on local disk at $8,229. Capacity and cost increased, but installed CPU fell because the selected hardware was more efficient. The narrower remaining risk belongs in opt-in model tests that hold family and storage constant and compare calculated `cpu_cores` requirements.

tests/conftest.py
```diff
@@ -1,36 +0,0 @@
-import pytest
-
-from service_capacity_modeling import capacity_planner
-from service_capacity_modeling.capacity_planner import CapacityPlanner
-from service_capacity_modeling.models.org import netflix
-
-
-@pytest.fixture(scope="session", autouse=True)
```
homatthew, line 8: This deletes an ineffective fixture, not a test. Test modules import `planner` directly during collection, before this session fixture replaces `capacity_planner.planner`, so their bound references continued using the production defaults. Expensive tests now declare the simulation count their own contract needs.

tests/netflix/test_counter.py
```diff
@@ -44,16 +44,15 @@ def test_eventual_counter_storage_targets():
         ),
     )
 
-    cap_plan = planner.plan(
+    cap_plan = planner.plan_certain(
```
homatthew, line 47: Counter chooses its composed storage targets before uncertainty sampling. The previous 256 simulations repeated the same routing decision; `plan_certain` exercises that decision directly without reducing the relevant input space.

tests/test_reproducible.py
```diff
@@ -45,12 +45,15 @@
 
 def test_repeated_plans():
     results = []
-    for _ in range(5):
+    # Two identical uncertain plans are sufficient to prove reproducibility.
+    # Keep both storage classes eligible so this covers the full Cassandra path.
+    for _ in range(2):
```
homatthew, line 50: Reproducibility requires two observations, not five. Both storage classes remain eligible, and the test still compares exact serialized plans. Four uncertainty samples are enough to expose nondeterminism without paying the 128-simulation convergence default.

tests/test_reproducible.py
```diff
@@ -76,12 +79,14 @@ def test_compositional():
         region="us-east-1",
         desires=uncertain_mid,
         num_results=4,
+        simulations=4,
```
homatthew, line 82: This is an identity contract, not a convergence test. Direct Cassandra planning and composed key-value planning receive the same four samples, then their Cassandra subplans and regrets are compared byte-for-byte. Increasing the sample count only repeats that equivalence.

tests/test_reproducible.py
```diff
@@ -146,40 +147,46 @@ def test_multiple_options_diversify_with_more_simulations():
...
-    arbitrary_large_number = 1024
-    assert arbitrary_small_number < arbitrary_large_number
+    num_results = 12
+    small_simulation_count = 12
+    large_simulation_count = 64
+    assert small_simulation_count < large_simulation_count
+
+    # This test protects sampling diversity across instance families. Fixing the
+    # storage class avoids multiplying that contract by unrelated EBS candidates.
+    model_arguments = {"require_local_disks": True}
```
homatthew, line 157: This test asks whether more samples expose more instance families. Attached storage is an orthogonal candidate dimension, so holding storage constant isolates that contract. Comparing 12 with 64 samples preserves the diversity assertion; 1,024 was not a meaningful boundary.

PR #329, opened 2026-09-11
Add hardware-only instance scale factors

arojas2003:
Adds scale_factors(from_instance, to_instance), a pure function of two hardware shapes answering "to preserve the capacity currently provisioned, how many to_instance do I need per from_instance?" It is for callers that must change instance type without a workload-aware plan available.

It is deliberately workload-blind and is not a capacity recommender: it preserves the existing provisioning decision, including any overprovisioning, so a hardware change does not silently alter a cluster's risk posture. Rightsizing stays a separate, deliberate decision.

Per-dimension factors (compute, memory, network, and local disk when the source shape has any) are exposed alongside the limiting dimension. The limiting dimension is the safe default, since whichever resource runs out first is the binding constraint; narrowing to a single dimension asserts headroom on the others, and is_limiting() lets a caller detect that its chosen dimension is not the one that binds.

Compute reuses `to_reference_cores()`, making it vCPU x clock x IPC scale. `cpu_ipc_scale` already folds in the hyperthreading factor, so comparing physical cores instead would double-count SMT and drop clock frequency.

Dimensions computed from a defaulted `cpu_ipc_scale` are flagged uncurated rather than raising, since 1.0 is correct for every shape that currently omits it. A source shape with local disk migrating to one without is reported as unrepresentable rather than absent.


service_capacity_modeling/models/instance_scale.py
```diff
@@ -0,0 +1,330 @@
+"""Hardware-only scale factors between two instance shapes.
```
jasonk000, line 1: Solid comments 👍

service_capacity_modeling/models/instance_scale.py
```diff
@@ -0,0 +1,330 @@
...
+          them preserves the current capacity
+        """
+        if self.to_capacity == 0:
+            if self.from_capacity == 0:
+                return 1.0
+            return float("inf")
+        return self.from_capacity / self.to_capacity
+
+    @property
+    def is_representable(self) -> bool:
```
jasonk000, line 112: It seems like an ARM -> x86 migration would do this too.

service_capacity_modeling/models/instance_scale.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @property
+    def compute(self) -> DimensionScale:
+        """Effective compute: vCPU count x clock x IPC scale.
+
+        Named "compute" rather than "cpu" on purpose: this is not a vCPU count.
+        Comparing vCPUs would make hyperthreaded and non-hyperthreaded shapes
+        look equivalent when they are not.
+        """
+        return self.dimensions[ResourceType.cpu]
```
jasonk000, line 193: Possibly should trap on known-bad changes, like arm->x86? I am not sure what the app semantics should be. I am not sure if it's a real scenario that might happen or just over-cautioun.

tests/test_instance_scale.py
```diff
@@ -0,0 +1,250 @@
...
+        assert factors.is_limiting(dimension)
+    assert factors.limiting.factor == 1.0
+
+
+def test_hyperthreaded_to_physical_cores() -> None:
+    """c5.2xlarge is 8 vCPU = 4 HT cores; c7a.2xlarge is 8 vCPU = 8 real cores.
+
+    cpu_ipc_scale already folds in both the architectural IPC uplift and the
+    1.5x non-SMT factor, so the compute factor must come from
+    vCPU x GHz x cpu_ipc_scale (11.83 -> 25.10, i.e. ~0.471).
```
jasonk000: I think this comment is liable to cause confusion in future if this model ever changes. Consider whether test is over-precise; I think the compute.factor > 0.4 is more appropriate than == here.

tests/test_instance_scale.py
```diff
@@ -0,0 +1,250 @@
...
+    """c5 -> c7a nearly doubles compute but keeps RAM identical at 15.26 GiB."""
+    factors = scale_factors("c5.2xlarge", "c7a.2xlarge")
+
+    assert factors.memory.factor == 1.0
+    assert factors.limiting.resource is ResourceType.mem_gib
+    assert factors.limiting_resource is ResourceType.mem_gib
+    assert factors.limiting_factor == 1.0
+    assert factors.is_limiting(factors.memory)
+
+    # The safety valve for a caller that narrowed to compute.
```
jasonk000, line 63: what does this mean

service_capacity_modeling/models/instance_scale.py
```diff
@@ -0,0 +1,330 @@
...
+    to_capacity: float
+    """Per-instance capacity of the shape being migrated to"""
+
+    curated: bool = True
+    """Whether every shape value feeding this dimension was explicitly set.
+
+    False means at least one side fell back to a pydantic default rather than a
+    curated value, so the factor may be wrong in a way that looks right. Today
+    this only ever happens for compute, via an unset ``cpu_ipc_scale``.
+    """
```
jasonk000: I would suggest either (1) provide a value to the caller, or (2) error. I don't think there is a state where `curated` is helpful to callers and leaks internal state.

tests/test_instance_scale.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def test_no_local_disk_omits_the_dimension() -> None:
+    factors = scale_factors("c5.2xlarge", "c7a.2xlarge")
+
+    assert factors.disk is None
+    assert ResourceType.disk_gib not in factors.dimensions
+
+
+def test_uncurated_ipc_scale_is_flagged_not_raised() -> None:
```
jasonk000: Likely drop this test OR replace it with an error handler if we are not OK to share the result.

jasonk000 (changes requested):
Directionally great, some tweaks to the API shape to make things easier for future. See notes.

service_capacity_modeling/models/instance_scale.py
arojas2003: As discussed, I'll return return an error if scm is missing any data (i.e. `cpu_ipc_scale` is not set)

tests/test_instance_scale.py
arojas2003: Agreed, I'll just keep the `compute.factor > 0.4` check

service_capacity_modeling/models/instance_scale.py
```diff
@@ -275,6 +256,11 @@ def scale_factors(
     Instance specs do not vary by region in SCM's data -- regions differ only in
     pricing and lifecycle -- so there is no region argument. Names are resolved
     through the global shape catalog and raise ``KeyError`` if unknown.
+
+    Raises ``ValueError`` if either shape has no curated ``cpu_ipc_scale``: an
```
jasonk000, line 260: I would make this more general: Raises `ValueError` if either shape cannot be reliably converted. Implementation can stay cpu-focused.

tests/test_instance_scale.py
```diff
@@ -110,26 +114,28 @@ def test_no_local_disk_omits_the_dimension() -> None:
...
-def test_uncurated_ipc_scale_is_flagged_not_raised() -> None:
-    """m5d.* ships no cpu_ipc_scale, so compute rests on the 1.0 default.
+def test_missing_cpu_ipc_scale_raises() -> None:
+    """m5d.* ships no cpu_ipc_scale, so compute would rest on the 1.0 default.
 
-    It happens to be right (Skylake is the 1.0 baseline), but it is un-asserted
-    data, so it is surfaced rather than silently folded in -- and rather than
-    raising, which would refuse a swap whose factor is in fact correct.
+    It happens to be right for m5d (Skylake is the 1.0 baseline), but this is
+    un-asserted data: silently trusting a default that's correct by
```
jasonk000, line 121: Can we tidy this up a bit, feels a bit slop-y.

jasonk000 (approved):
Small nits, but LGTM.

tests/test_instance_scale.py
arojas2003: Agreed. Cleaned up

service_capacity_modeling/models/instance_scale.py
arojas2003: The other fields are technically required by the ``Instance`` model.
However, I added the other checks anyway to not make it so mysterious to future readers why we are only checking ``cpu_ipc_scale``

tests/test_instance_scale.py
arojas2003: Replaced with an error handler

service_capacity_modeling/models/instance_scale.py
arojas2003: Yeah, I'll add an issue to the repo flagging this

